PR #4, opened 2015-09-09
Minor javadoc improvements

pveentjer:
The javadoc compiler gave variour warnings and causes noise. This
pr contains fixes for these issues. In java 8 it isn't allowed anymore
to use < p / > for empty lines.

At other places type information, return or exception information was
added.


RobAustin:
Thanks

PR #14, opened 2016-07-12
Support target throughput rates above one second in JLBH

### LINKED ISSUE #13: JLBH - Support throughput < 1 operation / s
I'd like to use JLBH for benchmarking bulk operations which may take more than 1 second to complete. In my case they take around  5 seconds, so want to specify a target throughput of less than 1/5 or 0.2 operations / second. However, the signature `JLBHOptions#throughput(int)` effectively prohibits target throughput rates of less than 1 operation / s. 

Are you open to support smaller target throughput rates? I could imagine an overloaded version with the signature `JLBHOptions#throughput(int, TimeUnit)` to specify the throughput per time unit and the default would be `TimeUnit.SECONDS`.

danielmitterdorfer:
Closes #13

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/Jvm.java
+++ b/src/main/java/net/openhft/chronicle/core/Jvm.java
@@ -166,10 +166,20 @@ public enum Jvm {
      * @param micros Time in micros
      */
     public static void busyWaitMicros(long micros) {
-        long waitUntil = System.nanoTime() + (micros * 1_000);
+        busyWaitUntil(System.nanoTime() + (micros * 1_000));
+    }
+
+    /**
+     * This method is designed tobe used when the time to be
+     * waited is very small, typically under a millisecond.
+     *
+     * @param waitUntil nanosecond precision counter value to await.
+     */
+    public static void busyWaitUntil(long waitUntil) {
         while (waitUntil > System.nanoTime()) {
         }
     }
+
 
     /**
      * Get the Field for a class by name.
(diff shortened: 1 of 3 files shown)


danielmitterdorfer:
@peter-lawrey I tried to keep the changeset as small as possible. However, I've seen that you busy wait in [`JLBH` when accounting for coordinated omission](https://github.com/OpenHFT/Chronicle-Core/blob/master/src/main/java/net/openhft/chronicle/core/jlbh/JLBH.java#L107-L109) and also in the [uncorrected case](https://github.com/OpenHFT/Chronicle-Core/blob/master/src/main/java/net/openhft/chronicle/core/jlbh/JLBH.java#L111). I was tempted to sleep for a while before busy waiting but this would have been a larger change than immediately necessary, so I wanted to discuss it separately. What do you think?

peter-lawrey:
We could add a check for delays of 2 ms or more, Jvm.pause(ms) 1
milli-second less and busy wait the rest. This way we get more accurate
timings but lower cpu utilisation for longer delays.
On 12 Jul 2016 3:05 PM, "Daniel Mitterdorfer" <notifications@github.com>
wrote:

@peter-lawrey <https://github.com/peter-lawrey> I tried to keep the
changeset as small as possible. However, I've seen that you busy wait in
JLBH when accounting for coordinated omission
<https://github.com/OpenHFT/Chronicle-Core/blob/master/src/main/java/net/openhft/chronicle/core/jlbh/JLBH.java#L107-L109>
and also in the uncorrected case
<https://github.com/OpenHFT/Chronicle-Core/blob/master/src/main/java/net/openhft/chronicle/core/jlbh/JLBH.java#L111>.
I was tempted to sleep for a while before busy waiting but this would have
been a larger change than immediately necessary, so I wanted to discuss it
separately. What do you think?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/OpenHFT/Chronicle-Core/pull/14#issuecomment-232057617>,
or mute the thread
<https://github.com/notifications/unsubscribe/ABBU8S9FCeXoSZXhd390o_xkTKVUSfA-ks5qU58SgaJpZM4JKbTP>
.

danielmitterdorfer:
That's what I also had in mind. I've implemented a mixed waiting strategy now.

danielmitterdorfer:
@peter-lawrey Did you already get a chance to look at my PR or did it get lost in Github notifications? :)

peter-lawrey:
I asked one of my guys to review this but forgot to merge.

danielmitterdorfer:
Great, thanks!

PR #24, opened 2016-12-08
Failing test for a bad interaction between StringInterner and hash64(StringBuilder)

scottkidder:


scottkidder:
Note: this is not my use case, but this test exhibits the same behavior.

PR #29, opened 2017-04-27
Fix map0 access pattern

sergebg:
Existing implementation does not follow the pattern, which can be found in FileChannelImpl implementations. As result in some cases it's possible to get OutOfMemoryError (map failed). Actually this error does not relate to any heap memory problem. And correct implementation tries to collect garbage (which is expected to contain mapped memory buffers) and retries map0 again. Finally, repeated OOME is transformed into IOException, which is again makes more sense.

In my case the problem happens when large queue is quickly scanned. And with higher probability this happens if I need to do some intensive random access reading by index. As result the process VmSize grows too quickly and proper gc() helps to save the application. 

So, there are 2 important points:
- follow map0 pattern
- do not terrify the application with OutOfMemoryError

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/OS.java
+++ b/src/main/java/net/openhft/chronicle/core/OS.java
@@ -16,6 +16,7 @@
 
 package net.openhft.chronicle.core;
 
+import net.openhft.chronicle.core.util.ThrowingFunction;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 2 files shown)


peter-lawrey:
This is a good fall back solution.  Chronicle Queue should release mapping deterministically, rather than using the GC to do it after the fact. If the GC is cleaning this up, it is a bug.   However Chronicle Queue might be used with another library which uses map, so this is a good choice all the same.

PR #31, opened 2017-06-09
ClassAliasPool: fix default ClassLoader

mstrap:
The default ClassLoader to be used when invoking
ClassAliasPool(ClassLookup parent) should be the class' own
ClassLoader which is most likely the same one for all
Chronicle JARs.

Previously, Thread.currentThread().getContextClassLoader() had been used
which may be different from the Chronicle classes' ClassLoader.
This may result in weird behavior, as reported by
OpenHFT/Chronicle-Map#127.


peter-lawrey:
Thank you.

PR #32, opened 2017-07-05
Add @Override to UnsafeMemory.allocateInstance

Pr0methean:
allocateInstance was previously missing the @Override annotation.


Pr0methean:
Reopened (deleted the wrong branch).

PR #39, opened 2017-09-18
Update pom.xml

MartyIX:
https://issues.apache.org/jira/projects/SUREFIRE/versions/12340297 - newer version of Apache Surefire plugin that fixes some more Java 9 issues


MartyIX:
There is also new Apache Maven Compiler plugin:

https://mvnrepository.com/artifact/org.apache.maven.plugins/maven-compiler-plugin/3.7.0

PR #41, opened 2017-10-10
Get correct MAX_DIRECT_MEMORY value in Java 9

MartyIX:
Hi,

this PR gets rid of the message `net.openhft.chronicle.core.Jvm: Unable to determine max direct memory` on Java 9.

Unfortunately, I don't really know if this patch has some far reaching consequences as I don't know the codebase that well.

Best regards,
Martin

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/Jvm.java
+++ b/src/main/java/net/openhft/chronicle/core/Jvm.java
@@ -58,11 +58,15 @@ public enum Jvm {
     @NotNull
     private static final ThreadLocalisedExceptionHandler DEBUG = new ThreadLocalisedExceptionHandler(Slf4jExceptionHandler.DEBUG);
 
-    private static final long MAX_DIRECT_MEMORY = maxDirectMemory0();
-    private static final int JVM_JAVA_MAJOR_VERSION = getMajorVersion0();
-    private static final boolean IS_JAVA_9_PLUS = JVM_JAVA_MAJOR_VERSION > 8;
+    private static final int JVM_JAVA_MAJOR_VERSION;
+    private static final boolean IS_JAVA_9_PLUS;
+    private static final long MAX_DIRECT_MEMORY;
 
     static {
+        JVM_JAVA_MAJOR_VERSION = getMajorVersion0();
+        IS_JAVA_9_PLUS = JVM_JAVA_MAJOR_VERSION > 8; // IS_JAVA_9_PLUS value is used in maxDirectMemory0 method.
+        MAX_DIRECT_MEMORY = maxDirectMemory0();
+
         try {
             bitsClass = Class.forName("java.nio.Bits");
             reservedMemory = bitsClass.getDeclaredField("reservedMemory");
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -58,9 +58,9 @@
     @NotNull
     private static final ThreadLocalisedExceptionHandler DEBUG = new ThreadLocalisedExceptionHandler(Slf4jExceptionHandler.DEBUG);
 
-    private static final long MAX_DIRECT_MEMORY = maxDirectMemory0();
     private static final int JVM_JAVA_MAJOR_VERSION = getMajorVersion0();
     private static final boolean IS_JAVA_9_PLUS = JVM_JAVA_MAJOR_VERSION > 8;
+    private static final long MAX_DIRECT_MEMORY = maxDirectMemory0();
```
MartyIX: This re-ordering is necessary as otherwise `maxDirectMemory0` uses incorrect `IS_JAVA_9_PLUS` value. This is really dangerous as some IDE changes order of fields automatically. So this is very error prone.

src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -406,7 +406,14 @@ public long usedDirectMemory() {
 
     private static long maxDirectMemory0() {
         try {
-            Class<?> clz = Class.forName("sun.misc.VM");
+            Class<?> clz;
+
+            if (IS_JAVA_9_PLUS) {
```
epickrram: If you're concerned about ordering, just call `getMajorVersion0()` and check the result. This method should only be invoked once, so it's not like there's going to be a large cost associated with it.

MartyIX: @epickrram There is another solution: Just move initialization of of static fields to `static {}` block.

The original code:

```
    private static final int JVM_JAVA_MAJOR_VERSION = getMajorVersion0();
    private static final boolean IS_JAVA_9_PLUS = JVM_JAVA_MAJOR_VERSION > 8;
    private static final long MAX_DIRECT_MEMORY = maxDirectMemory0();

    static {

```

Proposed code:

```
    private static final int JVM_JAVA_MAJOR_VERSION;
    private static final boolean IS_JAVA_9_PLUS;
    private static final long MAX_DIRECT_MEMORY;

    static {
        JVM_JAVA_MAJOR_VERSION = getMajorVersion0();
        IS_JAVA_9_PLUS = JVM_JAVA_MAJOR_VERSION > 8;
        MAX_DIRECT_MEMORY = maxDirectMemory0();
```

This way the risk of introducing a bug is lower.

epickrram: That would work. Please include a comment stating that the result of `IS_JAVA_9_PLUS` is used by `maxDirectMemory0`.

MartyIX: @epickrram Done

PR #68, opened 2018-06-07
Make chronicle-core run and tests pass on Java 11

snazy:
compilation requires Java 8

java.nio.Bits.reservedMemory was renamed to RESERVED_MEMORY in Java 11: Try to use 'reservedMemory' first, and fallback to 'RESERVED_MEMORY'


RobAustin:
Thanks

PR #69, opened 2018-06-08
Add a shutdown hook to shutdown referenceCleanerExecutor

longshorej:
I'm curious about the team's thoughts on this. This allows us to use ChronicleQueue in an OSGi-like environment ([landlord](https://github.com/huntc/landlord)).

Shouldn't have an impact for most users, but for us it helps greatly as we can ensure that all threads are terminated when we unload a classloader and call any shutdown hooks it registered.

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/util/WeakReferenceCleaner.java
+++ b/src/main/java/net/openhft/chronicle/core/util/WeakReferenceCleaner.java
@@ -12,6 +12,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/core/util/WeakReferenceCleaner.java
```diff
@@ -100,7 +105,7 @@ public void run() {
                     }
 
                 } catch (InterruptedException e) {
-                    LOGGER.warn("Interrupted while trying to retrieve reference, exiting.", e);
+                    LOGGER.debug("Interrupted while trying to retrieve reference, exiting.", e);
```
longshorej: This is an expected condition when the JVM shuts down, right? Hence moving to debug.

dpisklov:
@huntc the queue should be closed ideally in your environment during shutdown.

huntc:
We will need to remove the shutdown hook explicitly during queue closure then. 

If a shutdown hook isn’t ever removed then a ref to the class loader will be retained, and memory will not be freed.

PR #89, opened 2019-10-10
StringBuilder/String "coder" field reflection caching fix

sheinbergon:
These values are currently reassessed on each call, even though they are determined on JVM startup. Using the high-level interface in Chronicle, it can really slow down deserialization performance. (also, an exception is currently thrown and recovered from for each String/StringBuilder check, which is really bad for standard deserialziation operation).

This PR fixes both behaviors for pre and post JDK9 versions.


RobAustin:
thank you

PR #93, opened 2020-01-20
Added new field USER_HOME in OS class

asadganiev:
It is useful to use it instead of System.getProperty("user.home")
each time when the user stores config files in the home
dir in hidden folder like $USER_HOME/.Project (e.g. .gradle,
.IntellijIdea, .m2 etc)


dpisklov:
@asadganiev merged - thanks

PR #99, opened 2020-04-12
enhance: jdk14 map0 args changed

whitelilis:
from jdk14,   sun.nio.ch.FileChannelImpl.map0(int,long,long)   changed to  sun.nio.ch.FileChannelImpl.map0(int,long,long,boolean)


dpisklov:
Thanks for the PR - apparently I just merged another one with exact the same changes! So you can assume I merged yours ;)

PR #103, opened 2020-04-27
Improve ThreadDump by adding waiting method

### LINKED ISSUE #96: Improve ThreadDump by adding waiting method
Some tests using ThreadDump could be substantially sped up if there was a method that waited for all required threads to close. This would prevent tests from having to wait an arbitrary time before checking thread consistency potentially waiting unnecessarily long.

Menzani:
Closes #96


dpisklov:
Thanks!

PR #136, opened 2020-09-05
Add Instant constructor for SetTimeProvider

hpple:
Hi!

Sometimes it may happen that you already have some Instant timestamp to test with. E.g. Instant.now() or timestamp fetched from a test data source.

That would be handy to have a constructor to use it directly.


dpisklov:
ta

Issue #151, opened 2020-11-13
Add Null checking where applicable


minborg:
Methods annotated by `@NotNull` should produce a `NullPointerException` if invoked by a `null` argument. As of now, only intent is indicated.

minborg:
Methods annotated by `@NotNull` should produce a `NullPointerException` if invoked by a null argument. As of now, only intent is indicated.

JerryShea:
If you compile in intellij then it automatically does this instrumentation
btw.

My dream is that the compiler can do all this for us i.e. confirm at
compile time that there are no callers that pass null to a Non I'll arg.

JerryShea:
I believe the checkers framework is supposed to do this but I have not got
round to trying it recently

minborg:
I have filed: https://github.com/osundblad/intellij-annotations-instrumenter-maven-plugin/issues/53

JerryShea:
I'm not 100% convinced. I don't know if we should be weaving in extra stuff
to production code

minborg:
I understand your concerns @JerryShea 

However, I think that it is better to immediately throw an exception if the invariants of the method do not hold rather than detecting the problem at a later stage, in some unlikely cases even in another thread. The JIT compiler is able to inline the assertions if stacked. This is equivalent to using the ubiquitous `Objects::requireNonNull` which is widely seen as uncontroversial. 

The compiler still needs to do null checking under the hood so it is likely that the performance impact is negligible.

I think we should do some benchmarking once we have a working candidate up and running to confirm this before we merge this potential feature with the master branch.

minborg:
After some consideration, I agree with you @JerryShea . There are some potentially dangerous corner cases that may be affected by instrumentation. It is better to explicitly use `Objects.requireNonNull()` where applicable.

PR #154, opened 2020-11-21
Add net.openhft.chronicle.core.util.ObjectUtils.requireNonNull method

JerryShea:
Add net.openhft.chronicle.core.util.ObjectUtils.requireNonNull method
Add @NotNull to some method contracts and consistentise null checking in UnsafeMemory

Issue #151 has got me thinking... Null checking is a bugbear of mine.

What I am trying to do with this PR is to introduce a standard way of null checking which is optimally performant and also expresses intention with its name (and also its @NonNull annotation).

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/Memory.java
+++ b/src/main/java/net/openhft/chronicle/core/Memory.java
@@ -45,9 +45,9 @@ public interface Memory {
 
     void writeByte(long address, byte i8);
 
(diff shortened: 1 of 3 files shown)


peter-lawrey:
The concern has been that the intellij plugin triggers an IllegalArgumentException rather than a NullPointerException causing tests to fail if the plug in is enabled ie they expect an NPE.

PR #181, opened 2020-12-22
Jvm.fatal() -> Jvm.error() #180

JerryShea:


JerryShea:
As discussed with @peter-lawrey - error maps more closely to customer expectations

PR #199, opened 2021-02-05
Allow demo repos to be tracked

minborg:
This PR changes the behaviour of the Analytics module by introducing a white list with "contains" package names. Initially, it contains just one element: ".demo". Also, package names are not limited to three levels but are still distinct up to level three (e.g. `a.b.c.d` and `a.b.c.d.e` reports just one of them but the entire package name, for example, `a.b.c.d`).


minborg:
This PR is converted to a "Lazy-PR" meaning that unless there are any objections as per Feb 12, 24:00 BST, the PR will merge automatically.

PR #206, opened 2021-03-02
allow system properties loaded from system.properties file to be overriden from command line

JerryShea:
Right now if I have a property `foo=woo` set in system.properties and then define `-Dfoo=bar` on command line the value from command line is lost so System.getProperty("foo") == woo.

This change allows the -D on command line to override, which makes more sense from user's POV IMO.

This is not a big change but doing a PR as this is potentially surprising and impactive.


minborg:
I think this is good. It's the way Spring Boot et al. works.

PR #239, opened 2021-07-22
Add ability to check if the current thread is the background resource releaser thread

nicktindall:
Seems a bit less brittle than using the thread name? e.g. would be easy to switch to a multi-threaded BRR without affecting clients this way, or some other strategy


JerryShea (approved):

hft-team-city:
Released in [Chronicle-Core-2.21ea84](https://github.com/OpenHFT/Chronicle-Core/releases/chronicle-core-2.21ea84), [BOM-2.21ea207](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.21ea207)

PR #240, opened 2021-07-26
Implement priority shutdown hook to run them in controlled order

alamar:
Fixes ChronicleEnterprise/Chronicle-Map-Enterprise#57

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/io/BackgroundResourceReleaser.java
+++ b/src/main/java/net/openhft/chronicle/core/io/BackgroundResourceReleaser.java
@@ -1,6 +1,7 @@
 package net.openhft.chronicle.core.io;
 
 import net.openhft.chronicle.core.Jvm;
+import net.openhft.chronicle.core.shutdown.PriorityHook;
 
 import java.util.concurrent.ArrayBlockingQueue;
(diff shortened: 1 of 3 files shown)


nicktindall (approved):

src/main/java/net/openhft/chronicle/core/io/BackgroundResourceReleaser.java
```diff
@@ -19,6 +20,8 @@
     static {
         RELEASER.setDaemon(true);
         RELEASER.start();
+
+        PriorityHook.add(99, BackgroundResourceReleaser::releasePendingResources);
```
nicktindall, line 24: Hi @alamar, I was thinking the hook might stop the `RELEASER` thread gracefully, currently it interrupts and warns because I guess it doesn't know why it was interrupted, but the hook could signal to it to exit that loop (by setting a "running" flag to false maybe?) and it wouldn't need to warn because it knows it's terminating due to a shutdown.
The purpose of this thread is to continually run `performClose()` on any `AbstractCloseables` for which `shouldPerformCloseInBackground()` is true, so the thread probably needs to be one of the last things to terminate before a shutdown.

PR #248, opened 2021-09-15
Update JavaDoc on EventLoop interface

nicktindall:
To align with the change outlined in OpenHFT/Chronicle-Threads#93

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/threads/EventLoop.java
+++ b/src/main/java/net/openhft/chronicle/core/threads/EventLoop.java
@@ -46,15 +46,11 @@ public interface EventLoop extends Closeable {
     void unpause();
 
     /**
-     * Notify event loop to stop executing handlers. It is not expected that event loops can then be restarted
+     * Stop executing handlers and block until all handlers are complete.
+     * <p>
+     * It is not expected that event loops can then be restarted.
      */
     void stop();
-
-    /**
-     * @return {@code true} close has been called
-     */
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/core/threads/EventLoop.java
```diff
@@ -46,23 +46,19 @@
...
+     * <p>
+     * It is not expected that event loops can then be restarted.
      */
     void stop();
 
-    /**
-     * @return {@code true} close has been called
-     */
-    @Override
-    boolean isClosed();
```
nicktindall, line 57: I removed this because it's actually specified in QueryCloseable which is a parent class (perhaps it pre-dated QueryClosable?)

src/main/java/net/openhft/chronicle/core/threads/EventLoop.java
```diff
@@ -72,4 +68,12 @@
...
         return CleaningThread.inEventLoop(Thread.currentThread());
     }
+
+    /**
+     * Stop the event loop, then close any resources being held.
+     *
+     * Blocks until the handlers are all stopped and closed
+     */
+    @Override
+    void close();
```
nicktindall, line 78: Included this here despite it being in Closeable interface (hence the @Override annotation) so we could add that it will block until the loop is closed

JerryShea (approved):

PR #258, opened 2021-10-19
use an IdentitySet so as to avoid calling hashCode() when reference a…

JerryShea:
…ccounting https://github.com/ChronicleEnterprise/Chronicle-FIX/issues/838

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/io/AbstractCloseable.java
+++ b/src/main/java/net/openhft/chronicle/core/io/AbstractCloseable.java
@@ -30,6 +30,7 @@ import net.openhft.chronicle.core.util.WeakIdentityHashMap;
 import java.lang.reflect.Field;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.Set;
 import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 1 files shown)


nicktindall (approved):
Nice one!

JerryShea:
Everyone else please feel to add yourselves as reviewers

src/main/java/net/openhft/chronicle/core/io/AbstractCloseable.java
```diff
@@ -159,7 +160,7 @@ public static void assertCloseablesClosed() {
 
         synchronized (traceSet) {
             traceSet.removeIf(o -> o == null || o.isClosing());
-            Set<Closeable> nested = new HashSet<>();
+            Set<Closeable> nested = Collections.newSetFromMap(new IdentityHashMap<>());
             for (Closeable key : traceSet) {
                 addNested(nested, key, 1);
             }
```
nicktindall, line 166: I think the HashSet on the following line also needs to be an identity set

JerryShea: good call - thanks

minborg:
This was a very good catch! Well done!

alamar (approved):

PR #260, opened 2021-10-22
Add simple assertion of thread containment invariants, Fix #259

### LINKED ISSUE #259: Add a mechanism allowing simple assertion of thread containment invariants
This is needed in Bytes and other libraries.

minborg:
I have back-ported this solution from the Java17 Sandbox.

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/internal/util/NopThreadConfinementAsserter.java
+++ b/src/main/java/net/openhft/chronicle/core/internal/util/NopThreadConfinementAsserter.java
@@ -1,0 +1,12 @@
+package net.openhft.chronicle.core.internal.util;
+
+import net.openhft.chronicle.core.util.ThreadConfinementAsserter;
+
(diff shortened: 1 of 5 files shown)


RobAustin:
@minborg for the uneducated - can you explain exactly what this is and why it helps, in the issue/pull request comments.

RobAustin (approved):

minborg:
So, some classes (like a DTO class) fall into the category "thread-contained" which is (almost) the opposite of "thread-safe". 

A thread-safe object is characterized by the property that no combination of valid operations across threads exists such that the object can be placed in an illegal state. 

A thread-contained object does not have this property.  So, if we perform one operation (e.g. set a DTO property), then the operation of reading that property from another thread introduces the possibility that the object is observed to have an illegal state (e.g. reading partial writes of a long or not seeing an update at all). 

The `ThreadConfinementAsserter` is a first step in being able to detect unintended access across threads using zero-cost assertions. The asserter is expected to be expanded by additional methods such as the ability to explicitly hand over to another thread.

PR #261, opened 2021-10-22
rationalise UnsafeMemory.copyMemory functions, optimise performance, improve test coverage

JerryShea:
Confirmed now that this works with build-all. I had to restore the loop for some on heap Object copies

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/UnsafeMemory.java
+++ b/src/main/java/net/openhft/chronicle/core/UnsafeMemory.java
@@ -23,6 +23,7 @@ import net.openhft.chronicle.core.util.Ints;
 import net.openhft.chronicle.core.util.Longs;
 import net.openhft.chronicle.core.util.MisAlignedAssertionError;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import sun.misc.Unsafe;
 
 import java.lang.reflect.Field;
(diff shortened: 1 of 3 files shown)


RobAustin (approved):

rogersimmons (approved):

src/main/java/net/openhft/chronicle/core/UnsafeMemory.java
```diff
@@ -570,61 +547,63 @@ public void copyMemory(long fromAddress, long address, long length) {
...
-
+            UNSAFE.copyMemory(fromAddress, address, length);
         } else {
             copyMemory0(null, fromAddress, null, address, length);
         }
     }
 
     @Override
-    public void copyMemory(byte[] bytes, int offset, Object obj2, long offset2, int length) {
+    public void copyMemory(byte[] bytes, int offset, @Nullable Object obj2, long offset2, int length) {
```
nicktindall: I feel like the names `source` and `destination` and `sourceOffset` and `destinationOffset` might be clearer for all these methods.

nicktindall: (or similar)

nicktindall (approved):

src/main/java/net/openhft/chronicle/core/UnsafeMemory.java
JerryShea: yep - agree - I will change. I know you don't like short names, but how about src and dest?

minborg: I prefer `source` over `src` etc.

PR #277, opened 2021-11-19
Comment the use of test-framework until stable version is available

alamar:


glukos (approved):

RobAustin:
why this change ?

alamar:
@RobAustin chronicle-core does not depend on chronicle-bom and therefore can't refer to SNAPSHOT artifacts so we have to release test-framework firse.

minborg:
`chronicle-test-framework-1.22ea2` is officially released

PR #286, opened 2021-12-14
Remove security hotspots, Fix #285

### LINKED ISSUE #285: Remove security hotspots

minborg:


src/main/java/net/openhft/chronicle/core/util/ThreadIndexAssigner.java
```diff
@@ -44,6 +44,6 @@ public int getId() throws IllegalStateException, BufferOverflowException, Buffer
     }
 
     protected int nextIndex(int size) {
-        return ThreadLocalRandom.current().nextInt(size);
+        return new SecureRandom().nextInt(size);
```
nicktindall, line 47: Is it worth implementing a `ThreadLocalSecureRandom` (similar to `ThreadLocalRandom`), this seems like a sonar warning that will come up again and again? Also Core might be a good place to put it.

Or does something like that already exist in the standard library?

nicktindall (approved):

src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -1587,7 +1587,7 @@ public static String getCpuClass() {
                 if (Files.isReadable(path)) {
                     model = Files.lines(path)
                             .filter(line -> line.startsWith("model name"))
-                            .map(line -> line.replaceAll(".*: ", ""))
```
JerryShea, line 1590: does this code have tests?

src/main/java/net/openhft/chronicle/core/util/ThreadIndexAssigner.java
JerryShea: Or shall we just suppress the sonar warning? I don't see it provides much benefit - we don't use randoms for anything important, or attackable

minborg: In this case, there is no performance issue so I think we can stick with SecureRandom

src/main/java/net/openhft/chronicle/core/Jvm.java
minborg: No. We should break out this functionality into a separate method and create a test.

minborg: Added a test

PR #288, opened 2021-12-14
Issue/287

minborg:
This PR removes most issues flagged as Bugs in SQ

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/Jvm.java
+++ b/src/main/java/net/openhft/chronicle/core/Jvm.java
@@ -45,6 +45,8 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.*;
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -52,10 +54,11 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.LockSupport;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
-import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static java.lang.Runtime.getRuntime;
 import static java.lang.management.ManagementFactory.getRuntimeMXBean;
+import static java.util.stream.Collectors.*;
 import static net.openhft.chronicle.core.Bootstrap.*;
 import static net.openhft.chronicle.core.OS.*;
 import static net.openhft.chronicle.core.UnsafeMemory.UNSAFE;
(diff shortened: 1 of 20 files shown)


src/main/java/net/openhft/chronicle/core/internal/analytics/ReflectionUtil.java
```diff
@@ -46,7 +46,7 @@ public static Method methodOrThrow(@NotNull final String className,
             return analyticsClass.getMethod(methodName, parameterTypes);
         } catch (ReflectiveOperationException e) {
             Jvm.rethrow(e);
-            return null;
+            throw new IllegalStateException("We should never end up here");
```
nicktindall: I wonder if it's possible/worth making something like `Jvm.rethrowable(ex)` so you can do `throw Jvm.rethrowable(ex)` and not need to do this?

nicktindall: Or `throw Jvm.toUnchecked(ex)` might be more explicit

src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -1606,6 +1611,8 @@ public static String getCpuClass() {
                             Jvm.warn().on(CpuClass.class, "process " + cmd + " returned " + ret);
                     } catch (InterruptedException e) {
                         Jvm.warn().on(CpuClass.class, "process " + cmd + " waitFor threw ", e);
+                        // Restore the interrupt state...
+                        Thread.currentThread().interrupt();
```
JerryShea, line 1615: nice

src/main/java/net/openhft/chronicle/core/internal/analytics/ReflectionUtil.java
JerryShea: yeah the standard pattern is `throw Jvm.rethrow(e)` - we should do that and not `throw new IllegalStateException("We should never end up here")` as that loses `e`

src/main/java/net/openhft/chronicle/core/io/BackgroundResourceReleaser.java
```diff
@@ -98,7 +98,7 @@ public static void releasePendingResources() {
                     performRelease(o);
             }
             if (stopping)
-                RESOURCES.offer(POISON_PILL);
+                offerPoisonPill();
```
JerryShea: this changes behaviour. We don't want to warn if we can't add - I think these 2 changes should be reverted

src/main/java/net/openhft/chronicle/core/internal/analytics/ReflectionUtil.java
minborg: Fixed now

src/main/java/net/openhft/chronicle/core/io/BackgroundResourceReleaser.java
minborg: Added a flag for warning

JerryShea (approved):

PR #290, opened 2021-12-14
Remove most minor warnings, Fix #289

### LINKED ISSUE #289: Reduce the number of issues marked as minor

minborg:


src/main/java/net/openhft/chronicle/core/internal/analytics/ReflectionUtil.java
```diff
@@ -46,7 +46,7 @@ public static Method methodOrThrow(@NotNull final String className,
             return analyticsClass.getMethod(methodName, parameterTypes);
         } catch (ReflectiveOperationException e) {
             Jvm.rethrow(e);
-            return null;
+            throw new IllegalStateException("We cannot end up here");
```
nicktindall: again it would be a bit nicer on static analysis if we could avoid the need for this with something like `throw Jvm.toUnchecked(e)`

nicktindall (approved):

src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -1118,11 +1110,10 @@ public static ClassMetrics classMetrics(final Class<?> clazz) throws IllegalArgu
...
     private static ClassMetrics getClassMetrics(final Class<?> c) {
         final Class<?> superclass = c.getSuperclass();
-        int start = Integer.MAX_VALUE, end = 0;
+        int start = Integer.MAX_VALUE;
+        int end = 0;
         for (Field f : c.getDeclaredFields()) {
-            if ((f.getModifiers() & (Modifier.STATIC | Modifier.TRANSIENT)) != 0)
-                continue;
-            if (!f.getType().isPrimitive())
+            if ((f.getModifiers() & (Modifier.STATIC | Modifier.TRANSIENT)) != 0 || !f.getType().isPrimitive())
```
JerryShea, line 1122: I don't think it is worth making changes like these for a sonar minor warning

src/main/java/net/openhft/chronicle/core/Maths.java
```diff
@@ -164,7 +163,7 @@ public static double round1(double d) {
         if (d < 0)
             return -round1(-d);
         final double factor = 1e1;
-        if (!(d <= WHOLE_NUMBER / factor)) // to handle NaN
```
JerryShea: this changes behaviour - the comment explains why

nicktindall: Oooh good pickup

JerryShea (dismissed):
the expression simplification (!(d <= WHOLE_NUMBER / factor)) changes behaviour

The rest of the changes look OK but I don't think there is value in going after Sonar minor warnings

src/main/java/net/openhft/chronicle/core/Maths.java
JerryShea: We should add a test for NaN for one of these functions to capture this

minborg: Wow. This was clumsy of me to miss. Thanks, @JerryShea for saving my bacon...

PR #292, opened 2021-12-15
Issue/291

minborg:


src/test/java/net/openhft/chronicle/core/JvmTest.java
```diff
@@ -58,7 +58,7 @@ public void checkThreadDump() {
 
     @Test
     public void addToClassPath() {
-        Jvm.addToClassPath(JvmTest.class);
+        assertDoesNotThrow(() -> Jvm.addToClassPath(JvmTest.class));
```
JerryShea: why add assertDoesNotThrow? I can't see the point. If that code throws then the test will fail anyway

src/test/java/net/openhft/chronicle/core/JvmTest.java
```diff
@@ -122,8 +123,10 @@ public void testMaxDirectMemory() {
 
     @Test
     public void enableSignals() {
-        Jvm.signalHandler((Signal signal) -> System.out.println(signal + " occurred"));
-        Jvm.addSignalHandler((String signal) -> System.out.println(signal + " occurred"));
+        assertDoesNotThrow(() -> {
```
JerryShea: same comment as above

src/test/java/net/openhft/chronicle/core/MathsTest.java
```diff
@@ -350,42 +350,6 @@ public void longRunningRound() {
                 });
     }
 
-    @Test
-    @Ignore("Long running, avg score = 6879")
```
JerryShea, line 354: why take this out?

src/test/java/net/openhft/chronicle/core/UnsafeMemoryTest.java
```diff
@@ -79,102 +93,74 @@ public void readShort() {
...
         for (int i = 0; i <= 64; i++)
             if (onHeap == null) {
-                memory.writeOrderedInt(addr + i, 0);
-            } else {
-                memory.writeOrderedInt(object, addr + i, 0);
-            }
-    }
-
-    @Test
-    public void readInt() {
```
JerryShea, line 102: why remove?

src/test/java/net/openhft/chronicle/core/JvmTest.java
minborg: Each test should have at least one explicit assert. This explicitly shows the purpose of the test. 

Otherwise, it is hard to distinguish implicit assertions from omitted assertions. That is the main argument.

Sometimes more code that is clear is better than less which is less clear.

src/test/java/net/openhft/chronicle/core/UnsafeMemoryTest.java
minborg: Because the read and write tests are merged and actually asserts operations produce the correct effect.

JerryShea (approved):

PR #295, opened 2021-12-15
Reduce the number of issues marked as Critical, Fix #293

### LINKED ISSUE #293: Reduce the number of issues marked as Critical

minborg:


src/main/java/net/openhft/chronicle/core/UnsafeMemory.java
```diff
@@ -947,7 +959,7 @@ public void testAndSetInt(long address, long offset, int expected, int value) th
         if (UNSAFE.compareAndSwapInt(null, address, expected, value))
             return;
         int actual = UNSAFE.getIntVolatile(null, address);
-        throw new IllegalStateException("Cannot change at " + offset + " expected " + expected + " was " + actual);
+        throw new IllegalStateException(CANNOT_CHANGE_AT + offset + EXPECTED + expected + WAS + actual);
```
JerryShea, line 962: sorry but this kind of change does not add value

JerryShea (approved):

src/main/java/net/openhft/chronicle/core/UnsafeMemory.java
minborg: Let's unflag them in the future unless there is a very large number of repetitions.

PR #302, opened 2021-12-16
Issue/301

peter-lawrey:
This utility class will make it easier to support timed locks on resources, while checking if the thread involved has died or not.
In particular it will help simplify the logic for concurrent Named Tailers but should be useful elsewhere.
Note: it has practically 100% code coverage.

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
+++ b/src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
@@ -1,0 +1,131 @@
+package net.openhft.chronicle.core.threads;
+
+import net.openhft.chronicle.core.Jvm;
+import net.openhft.chronicle.core.OS;
+import net.openhft.chronicle.core.values.LongValue;
+
+import java.io.File;
+
+public class ThreadLock {
+    static final Metrics METRICS = new Metrics();
+    private static final long TOP_BITS = ~0L << 32;
+    private static final int UNLOCKED = 0;
+    private final LongValue twoThreadId;
+    private final long timeoutMs;
+    int busyLoopCount = 20_000;
+    int busyLockSlowerCount = 100;
+
+    public ThreadLock(LongValue twoThreadId, long timeoutMs) {
+        this.twoThreadId = twoThreadId;
+        this.timeoutMs = timeoutMs;
+    }
+
+    public boolean tryLock(int threadId) {
+        assert threadId != UNLOCKED;
+        long twoThreadId0 = twoThreadId.getVolatileValue();
+        int threadId0 = (int) twoThreadId0;
+        if (threadId0 != UNLOCKED) {
+            if (threadId0 == threadId)
+                throw new IllegalStateException("trying to lock twice");
+            return false;
+        }
+        return twoThreadId.compareAndSwapValue(twoThreadId0, (twoThreadId0 & TOP_BITS) | threadId);
+    }
+
+    public void lock(int threadId) throws InterruptedRuntimeException {
+        assert threadId != UNLOCKED;
+        if (tryLock(threadId))
+            return;
+        busyLock(threadId);
+    }
+
+    private void busyLock(int threadId) {
+        // busy wait
+        for (int i = 0; i < busyLoopCount; i++) {
+            if (tryLock(threadId))
+                return;
+            Jvm.nanoPause();
+        }
+        if (busyLockSlower(threadId))
+            return;
+        forceUnlockAndRetry(threadId);
+    }
+
+    private boolean busyLockSlower(int threadId) throws InterruptedRuntimeException {
+        long endMS = System.currentTimeMillis() + timeoutMs;
+        do {
+            for (int i = 0; i < busyLockSlowerCount; i++) {
+                if (tryLock(threadId))
+                    return true;
+                Thread.yield();
+            }
+            if (METRICS.supportsProc) {
+                int lockedThread = (int) twoThreadId.getVolatileValue();
+                if (lockedThread == 0)
+                    continue;
+                if (!isThreadRunning(lockedThread)) {
+                    Jvm.warn().on(getClass(), "ThreadId " + lockedThread + " died while holding a lock");
+                    return false;
+                }
+            }
+            if (Thread.currentThread().isInterrupted())
+                throw new InterruptedRuntimeException();
+
+        } while (System.currentTimeMillis() < endMS);
+        return false;
+    }
+
+    private boolean isThreadRunning(int lockedThread) {
+        return new File("/proc/" + lockedThread).exists();
+    }
+
+    private void forceUnlockAndRetry(int threadId) {
+        long twoThreadId0 = twoThreadId.getVolatileValue();
+        int threadId0 = (int) twoThreadId0;
+
+        // unlock the previous thread
+        if (twoThreadId.compareAndSwapValue(twoThreadId0, ((long) threadId0 << 32) | UNLOCKED)) {
+            String status = METRICS.supportsProc
+                    ? isThreadRunning(threadId0) ? "running" : "dead"
+                    : "unknown";
+            Jvm.warn().on(getClass(), "Successfully forced an unlock for threadId: " + threadId + ", previous thread held by: " + threadId0 + ", status: " + status);
+        } else {
+            Jvm.warn().on(getClass(), "Failed to forced an unlock for threadId: " + threadId);
+        }
+        lock(threadId);
+    }
+
+    public void unlock(int threadId) throws IllegalStateException {
+        long twoThreadId0 = twoThreadId.getVolatileValue();
+        int threadId0 = (int) twoThreadId0;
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
```diff
@@ -0,0 +1,134 @@
...
+            for (int i = 0; i < busyLockSlowerCount; i++) {
+                if (tryLock(threadId))
+                    return true;
+                Thread.yield();
+            }
+            if (METRICS.supportsProc) {
+                int lockedThread = (int) twoThreadId.getVolatileValue();
+                if (lockedThread == 0)
+                    continue;
+                if (!isThreadRunning(lockedThread)) {
```
rogersimmons, line 74: There appears to be a race condition here. 
If thread A holds the lock and is dead, and threads B and C try to acquire the lock, both B and C detect A is dead here.
B and C then both go to forceUnlockAndRetry. But forceUnlockAndRetry unconditionally clears the current owner.
So assuming B gets there first, B (correctly) clears A and acquires the lock
C then (incorrectly) clears B immediately after

To prevent this, busyLockSlower should return the "assumed dead" thread id to the caller. This is then explicitly passed to forceUnlockAndRetry - so forceUnlockAndRetry is told on entry which (dead) thread is believed to be holding the lock and needs to be killed, rather than reading the current value (which may now be a non-dead thread) at line 85.

rogersimmons (changes requested):

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
rogersimmons: Similarly on the timeout case: A is alive but times out, B and C both proceed to force unlock A. B unlocks A, but C then immediately unlocks B.

rogersimmons: (Strictly speaking, forceUnlockAndRetry should return the full twoThreadId to be force unlocked, so it can be used directly in the cas).

rogersimmons: Going further, rather than forceUnlock followed by a separate lock attempt, when passing in the known explicit dead/timedout twoThreadId both steps can be combined in a single atomic CAS: force-unlock-and-acquire by CASing on the old twoThreadId and setting the new value as (twoThreadID<<32 | newThreadId)

minborg: I suggest creating an interface for the intended functionality and then providing a static constructor method that will return an implementation located in the internal package.

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
```diff
@@ -0,0 +1,134 @@
...
+            }
+            if (Thread.currentThread().isInterrupted())
+                throw new InterruptedRuntimeException();
+
+        } while (System.currentTimeMillis() < endMS);
+        return false;
+    }
+
+    private boolean isThreadRunning(int lockedThread) {
+        return new File("/proc/" + lockedThread).exists();
```
minborg, line 87: Is there a way of doing this on platforms other than Linux?

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @SuppressWarnings("java:S3358")
```
minborg, line 90: Nähhee

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
```diff
@@ -0,0 +1,134 @@
...
+            if (prevThreadId == threadId)
+                throw new IllegalStateException("Lock already unlocked by threadId " + threadId);
+            else
+                Jvm.warn().on(getClass(), "Lock previously held by another thread " + prevThreadId + " not mine " + threadId);
+        } else {
+            Jvm.warn().on(getClass(), "Lock held by another thread " + threadId0 + " not mine " + threadId);
+        }
+    }
+
+    static class Metrics {
```
minborg, line 134: Is this class intended to be subclassed?

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
nicktindall, line 87: Where do we get the OS's thread ID from (such that we'd see a record for `/proc/{thread-id}`)?

The number we get from `Thread.getId()` seems to be synthetic/internal to the JVM (and counting up from 1 so not unique system-wide)

`Thread.getId()` implementation:
https://github.com/openjdk/jdk/blob/9a9add8825a040565051a09010b29b099c2e7d49/jdk/src/share/classes/java/lang/Thread.java#L1696-L1698

Initialization of `Thread#tid`
https://github.com/openjdk/jdk/blob/9a9add8825a040565051a09010b29b099c2e7d49/jdk/src/share/classes/java/lang/Thread.java#L421

Implementation of `Thread#nextThreadID()`:
https://github.com/openjdk/jdk/blob/9a9add8825a040565051a09010b29b099c2e7d49/jdk/src/share/classes/java/lang/Thread.java#L217-L219

nicktindall: I'm guessing we'd need to use some JNA/JNI trickery to get this?

src/test/java/net/openhft/chronicle/core/threads/ThreadLockTest.java
```diff
@@ -0,0 +1,267 @@
...
+        @Override
+        public long getValue() throws IllegalStateException {
+            return getValues.remove(0);
+        }
+
+        @Override
+        public void setValue(long value) throws IllegalStateException {
+            setValues.add(value);
+        }
+    }
```
nicktindall, line 266: Would be good to see a stress test that spun up a bunch of threads that contended for a lock and checked some guarded state to ensure multiple threads didn't get in. Including some that died holding the lock.

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
```diff
@@ -0,0 +1,134 @@
...
+            return;
+        }
+
+        if (twoThreadId.compareAndSwapValue(twoThreadId0, ((long) threadId << 32) | UNLOCKED))
+            return;
+        Jvm.warn().on(getClass(), "Failed to unlock");
+    }
+
+    private void unlock2(int threadId, long twoThreadId0, int threadId0) {
+        if (threadId0 == 0) {
```
nicktindall, line 123: Could we use the `UNLOCKED` constant here?

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
```diff
@@ -0,0 +1,134 @@
...
+            int prevThreadId = (int) (twoThreadId0 >>> 32);
+            if (prevThreadId == threadId)
+                throw new IllegalStateException("Lock already unlocked by threadId " + threadId);
```
nicktindall, line 126: Also, shouldn't unlock be idempotent? maybe log a warning? throwing seems extreme here?

src/test/java/net/openhft/chronicle/core/threads/ThreadLockTest.java
nicktindall: And a multi-process version might be important as well

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
peter-lawrey: I have changed forceLock to use a retained "old" lock value which timed out. If the value has changed in the meantime the force lock fails and it tries again.

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
peter-lawrey: It would be nice to know. Possibly a posix or system call.

peter-lawrey: PosixAPI.gettid() works on Linux.

src/test/java/net/openhft/chronicle/core/threads/ThreadLockTest.java
peter-lawrey: A test well worth adding. I did one performance test and it showed an issue.

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
peter-lawrey: It suggests a coding error to attempt to unlock multiple times. You have the option to catch the exception if you are just doing it as you are not sure if you have already unlocked it.

PR #304, opened 2021-12-20
Issue/303

minborg:


pom.xml
```diff
@@ -144,7 +144,7 @@
                         <configuration>
                             <referenceVersion>2.22ea3</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
-                            <binaryCompatibilityPercentageRequired>95.1</binaryCompatibilityPercentageRequired>
+                            <binaryCompatibilityPercentageRequired>92.6</binaryCompatibilityPercentageRequired>
```
minborg, line 147: I have manually checked the changes are ok. Mostly it is Enum converted to proper util classes meaning that values() etc. are missing. Also, internalized service implementations are safe to remove as they are not intended to be directly used.

src/main/java/net/openhft/chronicle/core/io/UnsafeText.java
```diff
@@ -58,7 +58,7 @@ public static long appendFixed(long address, long num) {
         return address;
     }
 
-    protected static void reverseTheOrder(long address, long start) {
+    private static void reverseTheOrder(long address, long start) {
```
minborg, line 61: The class was final so it is safe to make these private.

src/main/java/net/openhft/chronicle/core/onoes/ChainedExceptionHandler.java
```diff
@@ -19,18 +19,27 @@
...
+import static net.openhft.chronicle.core.util.ObjectUtils.requireNonNull;
 
 public class ChainedExceptionHandler implements ExceptionHandler {
     @NotNull
     private final ExceptionHandler[] chain;
 
     public ChainedExceptionHandler(@NotNull ExceptionHandler... chain) {
-        this.chain = chain;
+        requireNonNull(chain);
+        this.chain = Stream.of(chain)
```
minborg, line 36: Defensive copy

RobAustin (approved):

PR #307, opened 2021-12-22
Issue/301

peter-lawrey:


src/main/java/net/openhft/chronicle/core/internal/threads/VanillaThreadLock.java
```diff
@@ -0,0 +1,163 @@
...
+            if (tryLock(threadId))
+                return;
+            Jvm.nanoPause();
+        }
+        if (busyLockSlower(threadId))
+            return;
+        forceUnlockAndRetry(threadId);
+    }
+
+    @SuppressWarnings("java:S3776")
```
minborg, line 59: Rather than suppressing the warning for cognitive overload, please consider breaking up the logic into steps.

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
```diff
@@ -0,0 +1,30 @@
+package net.openhft.chronicle.core.threads;
+
+public interface ThreadLock {
```
minborg, line 3: +1 for providing an interface and an implementation in an internal package

src/main/java/net/openhft/chronicle/core/values/LongValue.java
```diff
@@ -75,4 +79,8 @@ default boolean isClosed() {
     @Override
     default void close() {
     }
+
+    default ThreadLock vanillaThreadLock() {
```
minborg, line 83: The name is silent whether or not a new instance is created on each invocation. Perhaps "createThreadLock()" would be more descriptive?

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Try to lock once or fail
+     *
+     * @param threadId to attempt to lock for
+     * @return true if the lock could be obtains, or false if not.
+     * @throws IllegalStateException if the same threadId already holds the lock
+     */
+    boolean tryLock(int threadId) throws IllegalStateException;
```
minborg, line 12: It is very unusual and generally discouraged to declare runtime exceptions in libraries. I think this is something we should move away from. There is already a JavaDoc tag for this here.

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
```diff
@@ -0,0 +1,30 @@
+package net.openhft.chronicle.core.threads;
+
+public interface ThreadLock {
+
+    /**
+     * Try to lock once or fail
+     *
+     * @param threadId to attempt to lock for
+     * @return true if the lock could be obtains, or false if not.
```
minborg, line 9: @return if the lock could be acquired

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
```diff
@@ -0,0 +1,30 @@
...
+    boolean tryLock(int threadId) throws IllegalStateException;
+
+    /**
+     * Lock the resource using a threadId
+     *
+     * @param threadId to lock
+     * @throws InterruptedRuntimeException if an interrupt occurred before or during a busy loop retry. It won't throw this if the lock can be obtained immediately.
+     * @throws IllegalStateException       if the same threadId already holds the lock
+     */
+    void lock(int threadId) throws InterruptedRuntimeException, IllegalStateException;
```
minborg, line 21: "@NonNegative int threadId"

minborg: Is the int value the lower 32 bits of the Thread.getId()?? In that case, it can be both negative and zero. Can we specify the requirements for what threadId is?

src/main/java/net/openhft/chronicle/core/threads/ThreadLock.java
minborg: Is a ThreadLock reentrant?

src/main/java/net/openhft/chronicle/core/values/LongValue.java
```diff
@@ -19,7 +19,11 @@
 package net.openhft.chronicle.core.values;
 
 import net.openhft.chronicle.core.Jvm;
+import net.openhft.chronicle.core.internal.threads.VanillaThreadLock;
 import net.openhft.chronicle.core.io.Closeable;
+import net.openhft.chronicle.core.threads.ThreadLock;
+
+import java.util.concurrent.TimeUnit;
 
 public interface LongValue extends Closeable {
```
minborg, line 28: The relation between a LongValue and a ThreadLock is unclear. Why are they related? What does the LongValue reflect etc.

PR #308, opened 2021-12-22
Issue/301

peter-lawrey:


src/main/java/net/openhft/chronicle/core/internal/threads/VanillaThreadLock.java
```diff
@@ -0,0 +1,165 @@
...
+    int busyLockSlowerCount = 100;
+    private long oldLock;
+
+    public VanillaThreadLock(LongValue twoThreadId, long timeoutMs) {
+        this.twoThreadId = twoThreadId;
+        this.timeoutMs = timeoutMs;
+    }
+
+    @Override
+    public boolean tryLock(int threadId) throws IllegalStateException {
```
nicktindall, line 27: I feel like we should call `PosixAPI.getid()` in here rather than taking the thread id as a parameter or otherwise we'll forever be taking support questions why it doesn't work when people wrongly pass `Thread.currentThread().getId()` in.

nicktindall: We should _at least_ rename the parameter to `posixThreadId` and describe where to get it from in the Javadoc

nicktindall: Maybe even check the value and warn if they pass in a value equal to `Thread.currentThread().getId()`, although that could be noisy in the event that value just happened to be equal to the posix thread ID.

JerryShea:
This PR replaces https://github.com/OpenHFT/Chronicle-Core/pull/307/ ? And https://github.com/OpenHFT/Chronicle-Core/pull/302 ?

benchmarks/pom.xml
```diff
@@ -36,7 +36,7 @@
                 <groupId>net.openhft</groupId>
                 <artifactId>third-party-bom</artifactId>
                 <type>pom</type>
-                <version>3.19.4</version>
+                <version>3.22.3</version>
```
RobAustin, line 39: consider changing to [3.22.3,)

src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -1596,13 +1625,6 @@ public void handle(final Signal signal) {
         }
     }
 
-    /**
-     * @return Obtain the model of CPU on Linux or the os.arch on other OSes.
-     */
-    public static String getCpuClass() {
```
RobAustin, line 1602: possibly a breaking API change!

src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -1682,27 +1704,4 @@ private CpuClass() {
             return line -> line.replaceFirst("[^:]*+: ", "");
         }
     }
-
-    /**
-     * Was assertion enabled for the Jvm class when it was initialised.
-     *
-     * @return if assertions were enabled.
-     */
-    public static boolean isAssertEnabled() {
```
RobAustin, line 1691: possibly a breaking API change!

src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -1682,27 +1704,4 @@ private CpuClass() {
...
-
-    public static boolean supportThread() {
-        String name = Thread.currentThread().getName();
-        return "Finalizer".equals(name) || name.contains("~");
-    }
-
-    /**
-     * park the current thread, and stay parked
-     */
-    public static void park() {
```
RobAustin, line 1703: possibly a breaking API change!

src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -1682,27 +1704,4 @@ private CpuClass() {
...
-    /**
-     * Was assertion enabled for the Jvm class when it was initialised.
-     *
-     * @return if assertions were enabled.
-     */
-    public static boolean isAssertEnabled() {
-        return ASSERT_ENABLED;
-    }
-
-    public static boolean supportThread() {
```
RobAustin, line 1695: possibly a breaking API change!

src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -1047,15 +1045,6 @@ public static void addSignalHandler(final SignalHandler signalHandler) {
     }
 
 
-    public interface SignalHandler {
```
RobAustin, line 1050: possibly a breaking API change!

RobAustin (approved):

src/main/java/net/openhft/chronicle/core/Jvm.java
peter-lawrey: This interfaces was moved to after the static methods

src/main/java/net/openhft/chronicle/core/Jvm.java
peter-lawrey: This static method was moved to before the non static members by the IDE

src/main/java/net/openhft/chronicle/core/Jvm.java
peter-lawrey: This static method was moved to before the non static members by the IDE

src/main/java/net/openhft/chronicle/core/Jvm.java
peter-lawrey: This static method was moved to before the non static members by the IDE

src/main/java/net/openhft/chronicle/core/Jvm.java
peter-lawrey: This static method was moved to before the non static members by the IDE

benchmarks/pom.xml
peter-lawrey: Is this comment out of date?

src/main/java/net/openhft/chronicle/core/internal/threads/VanillaThreadLock.java
peter-lawrey: Could double-check, however warnings can be easily ignored by developers.

peter-lawrey: I have added some convience methods to make it clearer how it should be used.

PR #310, opened 2021-12-22
Issue/309

minborg:


peter-lawrey (approved):

src/main/java/net/openhft/chronicle/core/util/NanoSampler.java
```diff
@@ -18,7 +18,21 @@
...
+import net.openhft.chronicle.core.annotation.SingleThreaded;
+
+@SingleThreaded
 @FunctionalInterface
 public interface NanoSampler {
+
+    /**
+     * Records a sample at the provided {@code nanos} time (corresponding to a time given in nanoseconds).
+     * <p>
+     * Provided nanos must be derived from a coherent and non-decreasing source such as {@link System#nanoTime()}
```
JerryShea, line 30: JLBH uses System.nanoTime internally so I think this can only be called with System.nanoTime  output

PR #312, opened 2021-12-22
Added support for nanosecond resolution Posix based wall clock time #311

peter-lawrey:


src/main/java/net/openhft/chronicle/core/time/PosixTimeProvider.java
```diff
@@ -0,0 +1,50 @@
...
+
+package net.openhft.chronicle.core.time;
+
+import net.openhft.posix.ClockId;
+import net.openhft.posix.PosixAPI;
+
+/**
+ * This is the raw Posix time, if you need it to be unique or a multiple of 32 use a wrapper.
+ */
+public enum PosixTimeProvider implements TimeProvider {
```
JerryShea, line 27: what is the difference between this and Chronicle-Ticker?

peter-lawrey: Good question, let be review it.

RobAustin: I've added @minborg for review as I think he suggests not using enums for utility methods like this.

RobAustin (approved):

src/main/java/net/openhft/chronicle/core/time/PosixTimeProvider.java
minborg: I think we should have a utility class that provides these different time providers and keep the actual implementations in an internal package. 

For example, `TimeProviders.acquirePosixTimeProvider()` or `TimeProviders.createUniqueMicroTimeProvider()`

minborg: `SetTimeProvider` is using a non-enum pattern which is better from an API perspective. (But not as good as the above TImeProviders.xxxxx

image

minborg: Are these classes to be used in conjunction with annotations? If not, we should really internalize them in the long run. We could move them to an internal package and deprecate the old ones (which could delegate form the internal ones and only act as a compatibility facade)

JerryShea: I thought we agreed at a dev meeting long ago we were going to continue to use enums for our singletons? We need to be consistent as right now we have different people doing different things

minborg: Enums for singletons are fine with me. But they are best put in an internal package and exposed by other means.

Issue #315, opened 2021-12-28
Create a tmp directory if one doesn't exists or is read only


peter-lawrey:
Some build systems have a read-only tmp directory (for the build agent)
For this reason, a `tmp` directory under TARGET should be created if needed.

hft-team-city:
Released in [Chronicle-Core-2.22ea31](https://github.com/OpenHFT/Chronicle-Core/releases/chronicle-core-2.22ea31), [BOM-2.22ea122](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.22ea122)

PR #316, opened 2021-12-28
Create a tmp directory if one doesn't exists or is read only #315

peter-lawrey:


src/main/java/net/openhft/chronicle/core/OS.java
```diff
@@ -115,6 +115,19 @@
     private OS() {
     }
 
+    private static String findTmp() {
+        String target = System.getProperty("project.build.directory");
+        if (target != null)
+            return target + File.pathSeparator + "tmp";
```
JerryShea, line 121: should we create the directory in this case?

src/main/java/net/openhft/chronicle/core/OS.java
```diff
@@ -115,6 +115,19 @@
...
+        final String tmp = System.getProperty("java.io.tmpdir");
+        if (tmp != null
+                && new File(tmp).isDirectory()
+                && new File(tmp).canWrite())
+            return tmp;
+        new File("tmp").mkdirs();
```
JerryShea, line 127: what if mkdirs() fails?

JerryShea (approved):

RobAustin (approved):

src/main/java/net/openhft/chronicle/core/OS.java
peter-lawrey: A subsequent IO operation that needs it will fail.

JerryShea: sonar will warn about ignoring the return value - maybe we should at the very least assert the return value?

PR #318, opened 2022-01-03
Deprecate ThreadLocalProperties, Fix #294

### LINKED ISSUE #294: Deprecate ThreadLocalProperties
Move this class to an internal package where it is used.

minborg:


JerryShea (approved):
Agree we should deprecate but don't we normally remove in current + 2 i.e. x.24?

RobAustin (approved):

Issue #320, opened 2022-01-05
Add a new method AbstractCloseable:assertCloseable


minborg:
Some resources that provide guarantees that certain operations will always succeed (eg Map that has an update lock) must be able to prevent resources from being closed. 

This cannot be implemented easily as `AbstractCloseable::close` will CAS its state variable regardless and cannot (rightfully) be overridden.

hft-team-city:
Released in [Chronicle-Core-2.22ea22](https://github.com/OpenHFT/Chronicle-Core/releases/chronicle-core-2.22ea22), [BOM-2.22ea79](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.22ea79)

PR #321, opened 2022-01-05
Issue/320

minborg:
Take a look at this PR. The object is to be able to prevent close() from being invoked in some cases. See #320

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/io/AbstractCloseable.java
+++ b/src/main/java/net/openhft/chronicle/core/io/AbstractCloseable.java
@@ -242,11 +242,9 @@ public abstract class AbstractCloseable implements ReferenceOwner, ManagedClosea
         return createdHere;
     }
 
-    /**
-     * Close a resource so it cannot be used again.
-     */
     @Override
     public final void close() {
+        assertCloseable();
         if (!UnsafeMemory.INSTANCE.compareAndSwapInt(this, CLOSED_OFFSET, STATE_NOT_CLOSED, STATE_CLOSING)) {
             if (shouldWaitForClosed() && isInUserThread()) {
                 waitForClosed();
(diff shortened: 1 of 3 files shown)


src/main/java/net/openhft/chronicle/core/io/AbstractCloseable.java
```diff
@@ -242,11 +242,9 @@ public StackTrace createdHere() {
         return createdHere;
     }
 
-    /**
```
RobAustin, line 245: why remove this javadoc ?

RobAustin (approved):

src/main/java/net/openhft/chronicle/core/io/AbstractCloseable.java
minborg: Because it was moved to the interface. This way, all implementations will be documented.

nicktindall (commented):
I can see the reasoning behind this, but it does concern me that we will most likely just ignore this exception and leave the resources open when it's thrown because we always use `closeQuietly`. I wonder if this is the correct approach?

It feels like it could end up being the source of flakiness or bugs.

minborg:
@nicktindall I agree that throwing an exception on close is not a perfect solution. If we take a step back and disregard the current solution, what would we ideally do when we are about to close a resource that is "acquired" by another thread? We could have the closing thread blocked until the resource is released, we could require close to acquiring an exclusive lock before actually closing, we could have a timer that expires after some time and only then throws an exception, etc.

minborg:
If we have a Map or a Queue and invoke close, then `closeQuietly()` will not be invoked.

nicktindall:
> @nicktindall I agree that throwing an exception on close is not a perfect solution. If we take a step back and disregard the current solution, what would we ideally do when we are about to close a resource that is "acquired" by another thread? We could have the closing thread blocked until the resource is released, we could require close to acquiring an exclusive lock before actually closing, we could have a timer that expires after some time and only then throws an exception, etc.

I don't know enough about the problem, but could you have each consumer of the resource take a lease and only do the actual close when the last lease was released? Like the reference counting? Would that make sense here?

JerryShea:
What is the requirement? We want a resource to be able to treat close as a request and veto the request to close?

minborg:
Yepp. The problem at hand was that Map used to throw an exception if someone else has acquired a lock and close() is invoked. The problem now is that when the new resource handler was introduced, it will first CAS on "closed" and then mask the exception thrown so no resource is released. But then we already CAS:ed the close operation so when a legit close tries to close a Map, it will silently ignore the close. Thus, we have a memory leak.

JerryShea:
So can we just fix this in Map and not have to change AbstractCloseable for everything? As @nicktindall has pointed out, throwing the exception will nearly always be ignored as our standard pattern is to call `Closeable.closeQuietly(...)`

minborg:
If we had used composition instead of inheritance, it would be an easy one. If anyone can come up with a good way of implementing a fix that is contained to Map only, I will be more than happy to implement it.

PR #346, opened 2022-01-27
Improve consistency of the Exception thrown by ClassLookup.forName(String name) #345

peter-lawrey:


pom.xml
```diff
@@ -29,7 +29,7 @@
       <sonar.host.url>https://sonarcloud.io</sonar.host.url>
     </properties>
     <artifactId>chronicle-core</artifactId>
-    <version>2.22ea27-SNAPSHOT</version>
+    <version>2.22ea28-SNAPSHOT</version>
```
JerryShea, line 32: suggest rebase to get rid of this

Issue #347, opened 2022-02-22
Added marker interface so the library knows an implementation will ignore everything


peter-lawrey:

hft-team-city:
Released in [Chronicle-Core-2.23ea0](https://github.com/OpenHFT/Chronicle-Core/releases/chronicle-core-2.23ea0), [BOM-2.22.129](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.22.129)

PR #348, opened 2022-02-22
Added marker interface so the library knows an implementation will ignore everything #347

peter-lawrey:


RobAustin (approved):

src/main/java/net/openhft/chronicle/core/Mocker.java
```diff
@@ -115,17 +117,20 @@ protected Object doInvoke(Object proxy, Method method, Object[] args) throws Inv
...
         addInterface(classes, tClass);
-        for (Class<?> aClass : additional) {
+        for (Class<?> aClass : additional)
             addInterface(classes, aClass);
-        }
+        classes.add(IgnoresEverything.class);
         try {
             //noinspection unchecked
-            return (T) Proxy.newProxyInstance(tClass.getClassLoader(), classes.toArray(NO_CLASSES), new AbstractInvocationHandler() {
+            final T t = (T) Proxy.newProxyInstance(ClassLoader.getSystemClassLoader(), classes.toArray(NO_CLASSES), new AbstractInvocationHandler() {
```
JerryShea, line 125: why change the class loader?

PR #350, opened 2022-02-23
Remove deprecated methods and classes in 2.23

alamar:
There's a failing test after the removal of `copyMemory()`, can you please look into it @JerryShea 
> net.openhft.chronicle.core.UnsafeMemory2Test#copyMemoryEachWayByteArrayHeapObject


src/main/java/net/openhft/chronicle/core/UnsafeMemory.java
```diff
@@ -566,20 +566,6 @@ public void copyMemory(long src, long dest, long length) {
         }
     }
 
-    @Override
-    public void copyMemory(byte[] src, int srcOffset, @Nullable Object dest, long destOffset, int length) {
```
alamar: One test fails after this overload is dropped

glukos (approved):

PR #355, opened 2022-03-10
ChronicleInit, migrate most system property access to Jvm methods

### LINKED ISSUE #354: Create ChronicleInit mechanism to set system properties before any of them are read by static initializers
This class should have a static initialization block (which may be replaced via "monkey patching") and should also be supplied via system property to provide an implementation of the earliest initialization code.

alamar:
closes #354

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/ChronicleInit.java
+++ b/src/main/java/net/openhft/chronicle/core/ChronicleInit.java
@@ -1,0 +1,36 @@
+package net.openhft.chronicle.core;
+
+/**
+ * Handles application code which must be loaded first/run and may override system properties.
+ *
+ * A {@link Runnable} fully qualified class name may be provided via {@code chronicle.init.runnable} property.
+ *
+ * This class may also be replaced with different concrete implementation. Code should reside in a static block to
+ * be run once. It should contain empty static init() method called to trigger class load.
+ */
+public final class ChronicleInit {
+    public static final String CHRONICLE_INIT_CLASS = "chronicle.init.runnable";
+
+    private ChronicleInit() {
+        // Suppresses default constructor, ensuring non-instantiability.
+    }
+
+    static {
+        // Jvm#getProperty() does not make sense here - not initialized yet
+        String initRunnableClass = System.getProperty(CHRONICLE_INIT_CLASS);
+        if (initRunnableClass != null && !initRunnableClass.isEmpty()) {
+            try {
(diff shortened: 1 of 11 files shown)


minborg:
I am not sure how to respond to this PR. What is the problem we are trying to solve?

As mentioned, we should not dig deeper into the rabbit hole of using static initializers. Instead, It would be much better to maintain instance variables that can be configured separately. This provides a much cleaner and better level of abstraction and allows, for example, parallel testing and prevents the many pitfalls of static initializers such as ordering, thread-safety, global state and deadlocks problems.

minborg:
Do we have to comb through all libraries and demos and replace `System.getProperty()` with `Jvm.getProperty()`?

src/main/java/net/openhft/chronicle/core/ChronicleInit.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Contains the application code which must be loaded first/run and may override system properties.
+ *
+ * A descendant of this class implementing {@link ChronicleInit#initialize()} method may be provided
+ * via {@code chronicle.init.class} property.
+ *
+ * This class may also be replaced with different concrete implementation. Code should reside in a static block to be run once.
+ * It should contain empty static init() method called to trigger class load.
+ */
+public abstract class ChronicleInit {
```
minborg: Providing an abstract class that is intended to be subclassed across package boundaries is highly undesirable as described in my article: https://dzone.com/articles/why-general-inheritance-is-flawed-and-how-to-final

It would be better to introduce an Interface that allows the injection of modifiers via the `ServiceLoader`

alamar:
@minborg please review updated PR

src/main/java/net/openhft/chronicle/core/ChronicleInit.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Handles application code which must be loaded first/run and may override system properties.
+ *
+ * A {@link Runnable} fully qualified class name may be provided via {@code chronicle.init.runnable} property.
+ *
+ * This class may also be replaced with different concrete implementation. Code should reside in a static block to
+ * be run once. It should contain empty static init() method called to trigger class load.
+ */
+public final class ChronicleInit {
```
minborg, line 11: Nice

alamar:
@minborg I've converted abstract class to Runnable, please take a look.

minborg:
> @minborg I've converted abstract class to Runnable, please take a look.

That's great!

minborg (approved):

PR #365, opened 2022-03-22
Add fallback System Properties for Greeting messages, Fix #364

### LINKED ISSUE #364: Add fallback System Properties for Greeting messages
GraalVM native images does not define the system properties:

* `java.runtime.name`
* `java.runtime.version`

We could add fallback for these under the keys:

* `java.vm.name`
* `java.vm.version`

minborg:


src/main/java/net/openhft/chronicle/core/internal/announcer/InternalAnnouncer.java
```diff
@@ -41,8 +41,8 @@ public static void announce(@NotNull final String groupId,
 
     private static void announceJvm() {
         LINE_PRINTER.accept(String.format("Running under %s %s with %d processors reported.",
-                Jvm.getProperty("java.runtime.name"),
-                Jvm.getProperty("java.runtime.version"),
+                systemPropertyOfEither("java.runtime.name", "java.vm.name"),
```
alamar: How about `Jvm.getProperty("java.runtime.name", Jvm.getProperty
("java.vm.name"))`?

minborg: I only want one of them. If `java.runtime.name` exists, then I want that corresponding value otherwise, I take `java.vm.name`'s value

alamar (approved):

PR #366, opened 2022-03-24
Remove JDK8ByteBufferCleanerService, Fix 363

minborg:
This PR removes `JDK8ByteBufferCleanerService` so JDK8 apps will use the reflective cleaner instead. The advantages are that GraalVM can build native images using the important flag `--allow-incomplete-classpath` and also reducing code size.

The performance impact would likely be negligible as we have static `MethodHandle` fields in the reflective cleaner service:

```
    private static final MethodHandle CLEANER_METHOD;
    private static final MethodHandle CLEAN_METHOD;
```


nicktindall (approved):

JerryShea (approved):
Sounds reasonable but would want @peter-lawrey to sign off too

minborg:
@peter-k-lawrey or @peter-lawrey  What is your opinion on the proposed change?

peter-lawrey (approved):

PR #368, opened 2022-03-30
SKIP_ASSERTIONS = false

minborg:
This is a temporary flick of this flag to make sure that all libraries build with the zero-cost assertions enabled. Ideally, this should be an automatic procedure during build all. Any ideas @alamar on how we could automate this process?

We should never release with this flag set.

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/util/AssertUtil.java
+++ b/src/main/java/net/openhft/chronicle/core/util/AssertUtil.java
@@ -126,7 +126,7 @@ public final class AssertUtil {
      * Performance critical code should use one of the first two schemes devised above to assert invariants.
      * The third, more convenient form, can be used for non-performance critical code.
      */
-    public static final boolean SKIP_ASSERTIONS = true;
+    public static final boolean SKIP_ASSERTIONS = false;
 
     // Suppresses default constructor, ensuring non-instantiability.
     private AssertUtil() {


src/main/java/net/openhft/chronicle/core/util/AssertUtil.java
```diff
@@ -126,7 +126,7 @@
      * Performance critical code should use one of the first two schemes devised above to assert invariants.
      * The third, more convenient form, can be used for non-performance critical code.
      */
-    public static final boolean SKIP_ASSERTIONS = true;
+    public static final boolean SKIP_ASSERTIONS = false;
```
alamar, line 129: Can we put it to system property? I can make sure all tests run with this property enabled.

alamar (changes requested):

src/main/java/net/openhft/chronicle/core/util/AssertUtil.java
alamar: Better yet, set it to Jvm.ASSERT_ENABLED by default, if no property is set

minborg: Unfortunately, that is not possible because then the compiler cannot collapse the expressions. But we might add a special JAR file before the normal JAR files that has only this class with `SKIP_ASSERTIONS = false;`. Classes found earlier on the CP JARs will take precedence.

minborg: Maybe a Maven profile?

minborg:
Another much better scheme is in the making so I will close this PR.

PR #369, opened 2022-03-31
Zeroassert

minborg:
This PR adds the possibility of selectively using zero-cost assertions in the build process.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,7 +61,7 @@
         <dependency>
             <groupId>net.openhft</groupId>
             <artifactId>${zero.cost.assertions}</artifactId>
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -55,6 +56,14 @@
     </dependencyManagement>
 
     <dependencies>
+
+        <!--TODO: Use profile to select type-->
```
alamar: Can you try doing that so we don't merge new feature with TODO?

minborg: Removed the comment as switching is done via -D parameters

alamar (approved):

PR #371, opened 2022-04-05
Make exception assertion logic reusable

nicktindall:
Just an idea, this code is repeated in all our `XXXTestCommon` classes, wouldn't it be easier/more consistent if it were packaged up for reuse like this?

I know this probably belongs in Chronicle-Test-Framework but `ExceptionKey` is in core, and core depends on Chronicle-Test-Framework so ideally we don't add a circular dependency there.

I made it release a test-jar so downstream projects could inherit it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,7 +113,6 @@
         <dependency>
             <groupId>net.openhft</groupId>
             <artifactId>chronicle-test-framework</artifactId>
-            <version>2.23ea2-SNAPSHOT</version>
             <scope>test</scope>
         </dependency>
 
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,8 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 
     <modelVersion>4.0.0</modelVersion>
 
@@ -26,8 +27,8 @@
     </parent>
     <properties>
       <zero.cost.assertions>disabled</zero.cost.assertions>
-      <sonar.organization>openhft</sonar.organization>
-      <sonar.host.url>https://sonarcloud.io</sonar.host.url>
+        <sonar.organization>openhft</sonar.organization>
+        <sonar.host.url>https://sonarcloud.io</sonar.host.url>
     </properties>
     <artifactId>chronicle-core</artifactId>
     <version>2.23ea9-SNAPSHOT</version>
@@ -209,6 +210,13 @@
                         </manifestEntries>
                     </archive>
                 </configuration>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
             </plugin>
 
             <!--
--- a/pom.xml
+++ b/pom.xml
@@ -113,6 +113,7 @@
         <dependency>
             <groupId>net.openhft</groupId>
             <artifactId>chronicle-test-framework</artifactId>
+            <version>2.23ea2-SNAPSHOT</version>
             <scope>test</scope>
         </dependency>
 
--- a/src/test/java/net/openhft/chronicle/core/CoreTestCommon.java
+++ b/src/test/java/net/openhft/chronicle/core/CoreTestCommon.java
@@ -1,25 +1,18 @@
 package net.openhft.chronicle.core;
 
+import net.openhft.chronicle.core.internal.ExceptionTracker;
 import net.openhft.chronicle.core.io.AbstractReferenceCounted;
-import net.openhft.chronicle.core.onoes.ExceptionKey;
-import net.openhft.chronicle.core.onoes.Slf4jExceptionHandler;
 import net.openhft.chronicle.core.threads.CleaningThread;
 import net.openhft.chronicle.core.threads.ThreadDump;
 import org.junit.After;
-import org.junit.Assert;
 import org.junit.Before;
-
-import java.util.LinkedHashMap;
-import java.util.Map;
-import java.util.function.Predicate;
 
 import static net.openhft.chronicle.core.io.AbstractCloseable.waitForCloseablesToClose;
 import static net.openhft.chronicle.core.io.AbstractReferenceCounted.assertReferencesReleased;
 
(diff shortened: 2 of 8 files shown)


minborg:
Looks very good to me. I agree that these features should reside in CTF as this would mean less library pollution but it would be a somewhat complicated and lengthy procedure to pull it off. `ExceptionKey` has a lot of improvement potential in it self. This is a cost/benefit issue. 

Could we define an interface (which is public) and then put the implementation in an internal package?

RobAustin (approved):

nicktindall:
@minborg I made an interface in CTF, and implemented it here, with a static factory method to make use of it. Do you think this gives us the abstraction we need to be able to evolve it? (see OpenHFT/Chronicle-Test-Framework#17)

pom.xml
```diff
@@ -112,6 +113,7 @@
         <dependency>
             <groupId>net.openhft</groupId>
             <artifactId>chronicle-test-framework</artifactId>
+            <version>2.23ea2-SNAPSHOT</version>
```
nicktindall: Just for the proposal

src/test/java/net/openhft/chronicle/core/internal/threads/VanillaThreadLockTest.java
```diff
@@ -93,7 +93,7 @@ public void lockTimeOut() {
         VanillaThreadLock lock = new VanillaThreadLock(value, 50);
         lock.lock(-1);
 
-        expectException("ThreadId -1 died while holding a lock");
+        ignoreException("ThreadId -1 died while holding a lock");
```
JerryShea, line 95: why did this change?

src/test/java/net/openhft/chronicle/core/CoreTestCommon.java
```diff
@@ -1,25 +1,19 @@
...
 import org.junit.Before;
 
-import java.util.LinkedHashMap;
-import java.util.Map;
-import java.util.function.Predicate;
-
 import static net.openhft.chronicle.core.io.AbstractCloseable.waitForCloseablesToClose;
 import static net.openhft.chronicle.core.io.AbstractReferenceCounted.assertReferencesReleased;
 
 public class CoreTestCommon {
```
JerryShea, line 14: I like this very much - this boilerplate is duplicated everywhere (sometimes slightly differently)

It's a bummer that ExceptionKey cannot be easily moved to test-framework. While *we* only use it for testing, I assume @peter-k-lawrey made it available for people to use in production code. @peter-k-lawrey is it worth reviewing this?

src/test/java/net/openhft/chronicle/core/internal/threads/VanillaThreadLockTest.java
nicktindall: So prior to this change `expectException` effectively meant `ignoreException`. I didn't change most of them (i.e. they have been upgraded to expectations), but some, the ones I changed in this PR didn't occur on either my machine or one of the other platform builds.

src/test/java/net/openhft/chronicle/core/CoreTestCommon.java
minborg: I agree @JerryShea . It would be highly desirable to move it to the test framework.

minborg:
This is really hard to get right...

src/test/java/net/openhft/chronicle/core/CoreTestCommon.java
```diff
@@ -1,25 +1,19 @@
 package net.openhft.chronicle.core;
 
+import net.openhft.chronicle.core.internal.JvmExceptionTracker;
 import net.openhft.chronicle.core.io.AbstractReferenceCounted;
-import net.openhft.chronicle.core.onoes.ExceptionKey;
-import net.openhft.chronicle.core.onoes.Slf4jExceptionHandler;
 import net.openhft.chronicle.core.threads.CleaningThread;
 import net.openhft.chronicle.core.threads.ThreadDump;
+import net.openhft.chronicle.testframework.internal.ExceptionTracker;
```
minborg, line 7: Later, the interface needs to be in the public API and not in `internal`.

src/test/java/net/openhft/chronicle/core/internal/JvmExceptionTracker.java
```diff
@@ -0,0 +1,93 @@
...
+     * Create a JvmExceptionTracker
+     *
+     * @return the exception tracker
+     */
+    public static JvmExceptionTracker create() {
+        return new JvmExceptionTracker(Jvm.recordExceptions());
+    }
+
+    /**
+     * {@inheritDoc}
```
minborg: JavaDocs are inherited by default so no need for `{@inheritDoc}` unless you add to the parent's method doc.

minborg:
Here is a way to decouple the implementation in Test from Core https://github.com/OpenHFT/Chronicle-Test-Framework/commit/a56104bb5b8c1cab68143e46d20f5665fc8c5883

It needs some more work but I run out of time...

alamar (approved):

src/test/java/net/openhft/chronicle/core/CoreTestCommon.java
nicktindall: Yeah I agree, but I was thinking we might "incubate" it for a release or so first, especially if we're deprecating or changing ExceptionKey and how all that's done.
Given it's not a feature we anticipate end users making a lot of use of I think it's safe to make it internal until we're happy it's settled?

nicktindall:
> Here is a way to decouple the implementation in Test from Core [OpenHFT/Chronicle-Test-Framework@a56104b](https://github.com/OpenHFT/Chronicle-Test-Framework/commit/a56104bb5b8c1cab68143e46d20f5665fc8c5883)
> 
> It needs some more work but I run out of time...

I finished it (I think)... it seems a little "overfit" to `Jvm`/`ExceptionKey`, but it'll do to start the process of moving that stuff to test framework. And it's internal so should be easy to evolve.

src/test/java/net/openhft/chronicle/core/internal/threads/VanillaThreadLockTest.java
JerryShea: I thought expectException meant "this exception will happen, ignore it, and throw if it does not happen" and ignoreException means "this exception may happen, ignore it"

nicktindall: Yeah that's what it means once you set it up that way (which I did in this PR), but in a lot of projects (including this one previously) expectException meant "this exception may happen, ignore it"

That's why I had to change some expects to ignores, because they either don't happen anywhere or don't happen on one of the platforms we test on

JerryShea: ah got it - nice

JerryShea (approved):

src/test/java/net/openhft/chronicle/core/CoreTestCommon.java
minborg: Just as we have an `internal` package we could have an `incubator` package. This is how Java itself is dealing with things like this: https://openjdk.java.net/jeps/11

PR #373, opened 2022-04-05
Catch the NoClassDefFoundError thrown when Windows/OSX attempts to look up a class with the wrong case

### LINKED ISSUE #372: Wrong exception type thrown when Windows/OSX attempts to look up an existing class with the wrong case
When Windows and OSX attempt to look up an existing class with the wrong case, they throw a `NoClassDefFoundError` instead of `ClassNotFoundException`

This means `ClassAliasPool#forName` throws an unexpected error in these circumstances

nicktindall:
Fixes #372

This fix was originally in Chronicle-Wire, but really it belongs here. I wanted to keep it as targeted to the specific case (on Windows/OSX only) as I could because I don't like the idea of catching Errors.

The only downside is we break the API for `ClassNotFoundRuntimeException`. If we don't want to do that inside a release I could make it only return the cause when it is actually a `ClassNotFoundException`?

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/pool/ClassAliasPool.java
+++ b/src/main/java/net/openhft/chronicle/core/pool/ClassAliasPool.java
@@ -19,6 +19,7 @@
 package net.openhft.chronicle.core.pool;
 
 import net.openhft.chronicle.core.Jvm;
+import net.openhft.chronicle.core.OS;
 import net.openhft.chronicle.core.util.ClassNotFoundRuntimeException;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/core/util/ClassNotFoundRuntimeException.java
```diff
@@ -1,12 +1,7 @@
 package net.openhft.chronicle.core.util;
 
 public class ClassNotFoundRuntimeException extends RuntimeException {
-    public ClassNotFoundRuntimeException(ClassNotFoundException cause) {
+    public ClassNotFoundRuntimeException(Throwable cause) {
         super(cause);
     }
-
-    @Override
-    public synchronized ClassNotFoundException getCause() {
```
JerryShea: ok to remove a public method without deprecating?

JerryShea (approved):

src/main/java/net/openhft/chronicle/core/util/ClassNotFoundRuntimeException.java
nicktindall: So to deprecate it, I could just return null when the cause wasn't a ClassNotFoundException (that should be exceedingly rare). And we could remove it at the next release (to return Throwable again).
If I don't deprecate it, and remove it, it breaks compilation in Wire. But that's the only place it's used (in our libraries) and it's only been around since February so I can't imagine it's heavily depended on outside of our org.
What do you think @peter-k-lawrey should we play it safe or go cowboy on this. We can simultaneously release Wire and Core so the break shouldn't be evident.

nicktindall: Scratch that I found a better way

PR #375, opened 2022-04-08
automatic activation of java11 & 17 profiles https://github.com/ChronicleEnterprise/RootPom/issues/218

JerryShea:


pom.xml
```diff
@@ -351,7 +351,11 @@
 
     <profiles>
         <profile>
+            <!-- this overrides java11 in java-parent-pom -->
             <id>java11</id>
+            <activation>
+                <jdk>[11,</jdk>
+            </activation>
```
nicktindall, line 357: possible conflict

nicktindall: It looks as though this is fine, see [here](http://www.dashbay.com/2011/03/maven-profile-inheritance/) apparently profiles are evaluated and applied prior to the child inheriting them. So profiles aren't inherited, but the effects of profiles are :exploding_head:

nicktindall: Which I think is a sensible design decision...

pom.xml
```diff
@@ -351,7 +351,11 @@
 
     <profiles>
         <profile>
+            <!-- this overrides java11 in java-parent-pom -->
```
nicktindall: it doesn't really (see that document I linked)

nicktindall (approved):

PR #377, opened 2022-04-11
fixed #376 added wget

### LINKED ISSUE #376: Added simple wget
see net.openhft.chronicle.core.io.Wget

RobAustin:


src/main/java/net/openhft/chronicle/core/io/Wget.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * performs and http get
+     *
+     * @param url the url of the http get
+     * @return the result, as a string
+     * @throws IOException if the connection could not be established
+     */
+    public static CharSequence url(String url) throws IOException {
+        final StringBuilder sb = stringBuilderPool.acquireStringBuilder();
```
alamar, line 28: Why we can't use URI to do the same?

src/main/java/net/openhft/chronicle/core/io/Wget.java
```diff
@@ -0,0 +1,46 @@
...
+        InputStream is = null;
+        try {
```
JanStureNielsen, line 30:
```
    public static CharSequence url(String url) throws IOException {
        StringBuilder builder = stringBuilderPool.acquireStringBuilder();

        try (InputStream input = URI.create(url).toURL().openStream()) {
            return builder.append(input.readAllBytes());
        }
    }
```

src/main/java/net/openhft/chronicle/core/io/Wget.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class Wget {
+
+    private static StringBuilderPool stringBuilderPool = new StringBuilderPool();
+
+    private Wget() {
+    }
+
+    /**
+     * performs and http get
```
JerryShea: "an"

src/main/java/net/openhft/chronicle/core/io/Wget.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.URL;
+
+import static net.openhft.chronicle.core.io.Closeable.*;
+
+public class Wget {
```
minborg: final

RobAustin: ok

src/main/java/net/openhft/chronicle/core/io/Wget.java
RobAustin: I agree what you have is more concise

```
    } finally {
            closeQuietly(is);
        }
``

but is missing the closeQuietly(s)

File: src/main/java/net/openhft/chronicle/core/io/Wget.java (hunk 1 above)
Comment (RobAustin) [reply src/main/java/net/openhft/chronicle/core/io/Wget.java:28]: @alamar I don't understand, do the same as what?

File: src/main/java/net/openhft/chronicle/core/io/Wget.java [hunk 5]
```diff
@@ -0,0 +1,46 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.URL;
+
+import static net.openhft.chronicle.core.io.Closeable.*;
+
+public final class Wget {
+
+    private static StringBuilderPool stringBuilderPool = new StringBuilderPool();
```
minborg, line 15: Suggest : `private static final StringBuilderPool STRING_BUILDER_POOL = new StringBuilderPool();`

src/main/java/net/openhft/chronicle/core/io/Wget.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private Wget() {
```
minborg, line 17: // Suppresses default constructor, ensuring non-instantiability.

src/main/java/net/openhft/chronicle/core/io/Wget.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * performs an http get
+     *
+     * @param url the url of the http get
+     * @return the result, as a string
+     * @throws IOException if the connection could not be established
+     */
+    public static CharSequence url(String url) throws IOException {
```
minborg, line 27: I think it would be better to provide a StringBuilder rather than magically get a TL on each invocation.  Maybe I pass the returned object around and then overwrite it on the next invocation.

minborg (approved):

PR #380, opened 2022-04-13
Make Bytes, BytesIn and ByteOut generic to fix raw type warning https://github.com/OpenHFT/Chronicle-Bytes/issues/358

JerryShea:


minborg (approved):
Smallest PR ever?

PR #383, opened 2022-04-13
Document support ticket 474. Fix #283.

### LINKED ISSUE #283: Document support ticket 474- BackgroundResourceReleaser
documenting new property `background.releaser.thread` as well as `runReleaseResources` and `stop` methods

foroughgoudarzi:


rogersimmons (changes requested):
It is always less than ideal when 2 closely related config options control a piece of functionality, due to the combinations which arise. On initial reading I found this description a bit convoluted - eg is it necessary to run BackgroundResourceReleaser.runBackgroundReleaserThread() if specifying the options? If so why both; if not why does the function exist?

Suggest a table may be better showing what each option turns on/off; and what still needs to be done manually with the options on/off respectively.

foroughgoudarzi:
@alamar this is the system property that you created. Could you please provide more information that addresses the above question so that I can document it.

rogersimmons (approved):

PR #385, opened 2022-04-20
Make default set of exception handlers explicit

### LINKED ISSUE #384: Jvm.resetExceptionHandlers() switches startup() handler from PERF to DEBUG
It is assumed that reset would return the exception handlers to what they were set to initially

nicktindall:
...so they can be restored when `Jvm.resetExceptionHandlers()` is called. Fixes #384

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/onoes/ThreadLocalisedExceptionHandler.java
+++ b/src/main/java/net/openhft/chronicle/core/onoes/ThreadLocalisedExceptionHandler.java
@@ -24,7 +24,15 @@ public class ThreadLocalisedExceptionHandler implements ExceptionHandler {
     private ExceptionHandler defaultHandler;
     private ThreadLocal<ExceptionHandler> handlerTL;
 
+    /**
+     * @deprecated use {@link ThreadLocalisedExceptionHandler(ExceptionHandler)} instead (to be removed in x.24)
+     */
+    @Deprecated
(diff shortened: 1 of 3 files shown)


src/main/java/net/openhft/chronicle/core/onoes/ThreadLocalisedExceptionHandler.java
```diff
@@ -24,7 +24,15 @@ public class ThreadLocalisedExceptionHandler implements ExceptionHandler {
     private ExceptionHandler defaultHandler;
     private ThreadLocal<ExceptionHandler> handlerTL;
 
+    /**
+     * @deprecated use {@link ThreadLocalisedExceptionHandler(ExceptionHandler)} instead (to be removed in .24)
+     */
+    @Deprecated
```
nicktindall, line 30: `binary-compatibility-enforcer` made me do it

JerryShea (approved):

src/main/java/net/openhft/chronicle/core/onoes/ThreadLocalisedExceptionHandler.java
JerryShea: hehe. minor but please use x.24 for easier greppage

minborg (commented):
Generally, we should move logger functionality from the Jvm class. It has nothing to do with the JVM and is a separate concern.

PR #391, opened 2022-05-02
Add @ScopeConfined annotation, Fix #390

### LINKED ISSUE #390: Add @ScopeConfined annotation
Add the following annotation:

```
/**
 * Indicates that the value, value consumer, value mapper etc. should not divulge the value outside
 * the annotated scope. The meaning value should also be construed to include any and all descendant Object properties
 * recursively from the original value.
 * <p>
 * This annotation can be used for entities that are dealing with reused-objects.
 * The term divulge obviously includes making the value or any of its descendant objects available to another Thread.
 *
 * @see ThreadConfined
 */
@Documented
@Retention(RetentionPolicy.CLASS)
@Target({ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE, ElementType.TYPE_USE})
public @interface ScopeConfined {

    /**
     * Returns a comment as to why this value must meet the condition.
     *
     * @return a comment as to why this value must meet the condition
     */
    String value() default "";
}
```

minborg:
```
/**
 * Indicates that the value consumer, value mapper, structure etc. should not divulge the object outside
 * the annotated scope. The meaning of "the object" should also be construed to include any and all descendant
 * non-immutable Object properties recursively dereferenced from the original object.
 * <p>
 * <em>
 * In other words,object contents would be purposely overwritten after they are passed to the receiver and
 * consequently cannot be kept after the invocation.
 * </em>
 * <p>
 * This annotation can be used for entities that are dealing with reused-objects.
 * The term divulge also includes making the value or any of its descendant objects available to another Thread.
 */
@Documented
@Retention(RetentionPolicy.RUNTIME)
@Target({ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE, ElementType.TYPE_USE})
public @interface ScopeConfined {

    /**
     * Returns a comment as to why this value must meet the condition.
     *
     * @return a comment as to why this value must meet the condition
     */
    String value() default "";
}
```

This is useful in case there are consumers/mappers etc. that operate on reused objects and we want to highlight this in the API more than just via the JavaDocs. 

Here is an example:

```
public interface HasIfPresent {

    /**
     * If a value is present, performs the given action with the value,
     * otherwise does nothing.
     * <p>
     * The value and its reference objects are {@code ScopeConfined} and <em>must only be used within the provided
     * {@code action}</em> and must not be leaked to the outside world.
     *
     * @param action the action to be performed, if a value is present
     * @throws NullPointerException if a value is present and the given action is
     *                              {@code null}
     */
    void ifPresent(Consumer<? super @ScopeConfined V> action);

}

```

Here is another example:

```
public interface HasStream<T> {

    /**
     * Returns a stream representing the current value if present, otherwise returns an
     * {@code empty()} stream.
     * <p>
     * The objects (if any) are {@code ScopeConfined} and should not be leaked outside the stream
     * (e.g. collecting etc.).
     *
     * @return a new object representing the current value if present, otherwise returns an
     * {@code empty()} object
     */
    Stream<@ScopeConfined T> stream();

}

```

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
+++ b/src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
@@ -21,17 +21,20 @@ package net.openhft.chronicle.core.annotation;
 import java.lang.annotation.*;
 
 /**
- * Indicates that the value, value consumer, value mapper etc. should not divulge the value outside
- * the annotated scope. The meaning value should also be construed to include any and all descendant Object properties
- * recursively from the original value.
+ * Indicates that the value consumer, value mapper, structure etc. should not divulge the object outside
+ * the annotated scope. The meaning of "the object" should also be construed to include any and all descendant
+ * non-immutable Object properties recursively dereferenced from the original object.
+ * <p>
+ * <em>
+ * In other words,object contents would be purposely overwritten after they are passed to the receiver and
+ * consequently cannot be kept after the invocation.
+ * </em>
  * <p>
  * This annotation can be used for entities that are dealing with reused-objects.
- * The term divulge obviously includes making the value or any of its descendant objects available to another Thread.
- *
- * @see ThreadConfined
+ * The term divulge also includes making the value or any of its descendant objects available to another Thread.
  */
 @Documented
-@Retention(RetentionPolicy.CLASS)
+@Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE, ElementType.TYPE_USE})
 public @interface ScopeConfined {
 
--- a/src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
+++ b/src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2016-2020 chronicle.software
+ *
+ * https://chronicle.software
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.openhft.chronicle.core.annotation;
+
+import java.lang.annotation.*;
+
+/**
+ * Indicates that the value, value consumer, value mapper etc. should not divulge the value outside
+ * the annotated scope. The meaning value should also be construed to include any and all descendant Object properties
+ * recursively from the original value.
+ * <p>
+ * This annotation can be used for entities that are dealing with reused-objects.
+ * The term divulge obviously includes making the value or any of its descendant objects available to another Thread.
+ *
+ * @see ThreadConfined
+ */
+@Documented
+@Retention(RetentionPolicy.CLASS)
+@Target({ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE, ElementType.TYPE_USE})
+public @interface ScopeConfined {
+
+    /**
+     * Returns a comment as to why this value must meet the condition.
+     *
+     * @return a comment as to why this value must meet the condition
+     */
+    String value() default "";
+}
--- a/src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
+++ b/src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
@@ -21,17 +21,17 @@ package net.openhft.chronicle.core.annotation;
 import java.lang.annotation.*;
 
 /**
- * Indicates that the value consumer, value mapper, structure etc. should not divulge the object outside
+ * Indicates that the value consumer, value mapper, structure etc. should not make the object available outside
  * the annotated scope. The meaning of "the object" should also be construed to include any and all descendant
  * non-immutable Object properties recursively dereferenced from the original object.
  * <p>
  * <em>
- * In other words,object contents would be purposely overwritten after they are passed to the receiver and
+ * In other words, object contents would be purposely overwritten after they are passed to the receiver and
  * consequently cannot be kept after the invocation.
  * </em>
  * <p>
  * This annotation can be used for entities that are dealing with reused-objects.
- * The term divulge also includes making the value or any of its descendant objects available to another Thread.
+ * The term make available also includes making the object or any of its descendant objects available to another Thread.
  */
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
--- a/src/test/java/net/openhft/chronicle/core/annotation/ScopeConfinedTest.java
+++ b/src/test/java/net/openhft/chronicle/core/annotation/ScopeConfinedTest.java
@@ -1,0 +1,33 @@
+package net.openhft.chronicle.core.annotation;
+
+import org.junit.jupiter.api.Test;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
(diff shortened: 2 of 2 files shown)


alamar:
Do we have some description of this mechanism (maybe it's a common practice and there are any intros?) Can you add a scoping of Core in the same commit along with the annotation?

src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+package net.openhft.chronicle.core.annotation;
+
+import java.lang.annotation.*;
+
+/**
+ * Indicates that the value consumer, value mapper, structure etc. should not divulge the object outside
+ * the annotated scope. The meaning of "the object" should also be construed to include any and all descendant
```
alamar, line 25: I had to look some of these up in a dictionary

src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE, ElementType.TYPE_USE})
+public @interface ScopeConfined {
+
+    /**
+     * Returns a comment as to why this value must meet the condition.
+     *
+     * @return a comment as to why this value must meet the condition
```
alamar, line 44: How does "V" answer the question why the value must meet the condition?

src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.lang.annotation.*;
+
+/**
+ * Indicates that the value consumer, value mapper, structure etc. should not divulge the object outside
+ * the annotated scope. The meaning of "the object" should also be construed to include any and all descendant
+ * non-immutable Object properties recursively dereferenced from the original object.
+ * <p>
+ * <em>
+ * In other words,object contents would be purposely overwritten after they are passed to the receiver and
```
alamar: White space is required after comma

src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
```diff
@@ -0,0 +1,47 @@
...
+ * consequently cannot be kept after the invocation.
+ * </em>
+ * <p>
+ * This annotation can be used for entities that are dealing with reused-objects.
+ * The term divulge also includes making the value or any of its descendant objects available to another Thread.
```
alamar: Maybe just use "keep" as a word?

src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
alamar: Or we're not doing it anymore...

minborg: > Or we're not doing it anymore...

Correct. Now the annotation would be pinned to the V itself.

src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
vach-chronicle: same, but i love it )

src/test/java/net/openhft/chronicle/core/annotation/ScopeConfinedTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+class ScopeConfinedTest {
+
+    @Test
+    void a() throws NoSuchMethodException {
+        Method method = Foo.class.getMethod("stream");
+        final Type genericReturnType = method.getGenericReturnType();
+
+        // Not sure how to get the Annotation...
+        assertEquals("java.util.stream.Stream<T>", genericReturnType.getTypeName());
```
alamar, line 20: You can get annotations via reflection, all right.

src/test/java/net/openhft/chronicle/core/annotation/ScopeConfinedTest.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.reflect.Type;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class ScopeConfinedTest {
+
+    @Test
+    void a() throws NoSuchMethodException {
```
alamar, line 15: I would expect descriptive name

src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
```diff
@@ -0,0 +1,47 @@
...
+ * consequently cannot be kept after the invocation.
+ * </em>
+ * <p>
+ * This annotation can be used for entities that are dealing with reused-objects.
+ * The term make available also includes making the object or any of its descendant objects available to another Thread.
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE, ElementType.TYPE_USE})
+public @interface ScopeConfined {
```
alamar, line 39: Annotation name is still confusing in my opinion.

nicktindall:
The thing I don't quite get is the annotation on generic parameters e.g. from the examples...
```
void ifPresent(Consumer<? super @ScopeConfined V> action);
```
Do we do this to allow standard interfaces like Stream/Consumer to have `@ScopeConfined` semantics applied?

Does this mean: all parameters of type V in this interface must be annotated as scope confined for consumers passed to `ifPresent(...)`

i.e. any consumer that is passed in must be of the form...
```
aThing.ifPresent((@ScopeConfined SomeObject so) -> {
   ...
});
```
That makes sense I guess. Would it be a compile error if the passed consumer didn't have the annotation?

It's a bit of a blunt instrument perhaps when there are multiple parameters of that type. e.g.
```
interface SomeInterface<V> {

    V doSomething(V firstInput, int anIntValue, V secondInput);
}
```
`class ScopeConfinedSomeInterface implements SomeInterface<? super @ScopeConfined V>` would apply to all V parameters & return types in the interface.

I'm guessing if you wanted to have finer grained control you'd make a specific interface instead of annotating type parameters?

src/test/java/net/openhft/chronicle/core/annotation/ScopeConfinedTest.java
minborg: I think types are erased at runtime but please give it a go.

src/main/java/net/openhft/chronicle/core/annotation/ScopeConfined.java
minborg: Any suggestions out there?

minborg:
> The thing I don't quite get is the annotation on generic parameters e.g. from the examples...
> 
> ```
> void ifPresent(Consumer<? super @ScopeConfined V> action);
> ```
> 
> Do we do this to allow standard interfaces like Stream/Consumer to have `@ScopeConfined` semantics applied?
> 
> Does this mean: all parameters of type V in this interface must be annotated as scope confined for consumers passed to `ifPresent(...)`
> 
> i.e. any consumer that is passed in must be of the form...
> 
> ```
> aThing.ifPresent((@ScopeConfined SomeObject so) -> {
>    ...
> });
> ```
> 
> That makes sense I guess. Would it be a compile error if the passed consumer didn't have the annotation?
> 
> It's a bit of a blunt instrument perhaps when there are multiple parameters of that type. e.g.
> 
> ```
> interface SomeInterface<V> {
> 
>     V doSomething(V firstInput, int anIntValue, V secondInput);
> }
> ```
> 
> `class ScopeConfinedSomeInterface implements SomeInterface<? super @ScopeConfined V>` would apply to all V parameters & return types in the interface.
> 
> I'm guessing if you wanted to have finer grained control you'd make a specific interface instead of annotating type parameters?

There is no requirement that Consumers et al. declare `@ScopeConfined` just as the case is for `@NotNull` annotations. The purpose is to have a standard way of expressing this (common) property instead of having to do extensive documentation for a large number of structures. 

Having specific interfaces breaks compatibility with some existing ecosystems like Stream, Consumer etc.

In short, this annotation indicates that the parameter should not be retained outside its scope. No enforcement is made and we might drop this proposal and instead copy the text from the annotation into a large number of method JavaDocs.

RobAustin (approved):

PR #395, opened 2022-05-09
Add sortpom plugin

minborg:
Added the "sortpom" plugin that keeps the POM neatly sorted in a specific way.

We could add this plugin to other libraries but let's start with this one and evaluate.


nicktindall (approved):
My opinions on re-ordering code without good reason are well known, but I'd make a reluctant approval on this because it's only POM files and we rarely need to do archaeology/cherry-picks on them.

PR #400, opened 2022-06-01
Remove junit-jupiter-engine from test classpath

nicktindall:
Get junit4 off the classpath, we don't use it anymore

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,8 +25,8 @@
         <relativePath></relativePath>
     </parent>
     <artifactId>chronicle-core</artifactId>
-    <packaging>bundle</packaging>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -25,8 +25,8 @@
         <relativePath></relativePath>
     </parent>
     <artifactId>chronicle-core</artifactId>
-    <packaging>bundle</packaging>
     <version>2.23ea13-SNAPSHOT</version>
+    <packaging>bundle</packaging>
```
nicktindall, line 29: I think sortpom does this?

JerryShea (approved):

PR #404, opened 2022-06-06
Issue/342

minborg:
This PR adds around 500 dynamic tests that assert the visibility and correctness of many atomic operations.

Also, some additional convenience methods for alignment are added.

### DIFF
--- a/src/test/java/net/openhft/chronicle/core/UnsafeMemoryLongTest.java
+++ b/src/test/java/net/openhft/chronicle/core/UnsafeMemoryLongTest.java
@@ -1,0 +1,98 @@
+package net.openhft.chronicle.core;
+
+import net.openhft.chronicle.core.util.Ints;
+import net.openhft.chronicle.testframework.Series;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.IntPredicate;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
+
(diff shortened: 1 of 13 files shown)


src/test/java/net/openhft/chronicle/core/UnsafeMemoryLongTest.java
```diff
@@ -0,0 +1,98 @@
...
+                new NamedOperation<>("UnsafeMemory::unsafeGetLong", (m, a) -> UnsafeMemory.unsafeGetLong(a)),
+                new NamedOperation<>("UnsafeMemory::readLong", UnsafeMemory::readLong),
+                new NamedOperation<>("UnsafeMemory::readVolatileLong", UnsafeMemory::readVolatileLong));
+    }
+
+    @Override
+    public List<NamedOperation<MemoryObjLongObjConsumer<Long>>> objectWriteOperations() {
+        return Arrays.asList(
+                new NamedOperation<>("UnsafeMemory::unsafePutLong(Object)", (m, obj, offset, v) -> UnsafeMemory.unsafePutLong(obj, offset, v)),
+                // SIGSEGV
```
alamar: Why?

minborg: So, this operation does not support `obj == null` mode like most of the others. I have documented this now.

alamar (approved):

Issue #409, opened 2022-06-09
Added SingleThreadedChecked to make resetting and disabling this check more consistent


peter-lawrey:

hft-team-city:
Released in [Chronicle-Core-2.23ea20](https://github.com/OpenHFT/Chronicle-Core/releases/chronicle-core-2.23ea20), [BOM-2.23ea72](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23ea72)

PR #410, opened 2022-06-09
Added SingleThreadedChecked to make resetting and disabling this chec…

peter-lawrey:
…k more consistent #409

Note, the previous methods returned `this` but getting this right required breaking changes to the generic of parents etc.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,10 +22,10 @@
         <groupId>net.openhft</groupId>
         <artifactId>java-parent-pom</artifactId>
         <version>1.1.33</version>
-        <relativePath />
+        <relativePath></relativePath>
     </parent>
     <artifactId>chronicle-core</artifactId>
-    <version>2.23ea14-SNAPSHOT</version>
+    <version>2.23ea20-SNAPSHOT</version>
     <packaging>bundle</packaging>
     <name>OpenHFT/Chronicle-Core</name>
     <description>Chronicle-Core</description>
@@ -192,7 +192,7 @@
                         <configuration>
                             <referenceVersion>2.23ea0</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
-                            <binaryCompatibilityPercentageRequired>100.0</binaryCompatibilityPercentageRequired>
+                            <!--                            <binaryCompatibilityPercentageRequired>99</binaryCompatibilityPercentageRequired>-->
                         </configuration>
                     </execution>
                 </executions>
--- a/src/main/java/net/openhft/chronicle/core/io/AbstractReferenceCounted.java
+++ b/src/main/java/net/openhft/chronicle/core/io/AbstractReferenceCounted.java
@@ -16,9 +16,11 @@ public abstract class AbstractReferenceCounted implements ReferenceCountedTracer
     protected static final long WARN_NS = (long) (Jvm.getDouble("reference.warn.secs", 0.003) * 1e9);
     protected static final int WARN_COUNT = Jvm.getInteger("reference.warn.count", Integer.MAX_VALUE);
     static volatile Set<AbstractReferenceCounted> referenceCountedSet;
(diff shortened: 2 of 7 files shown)


pom.xml
```diff
@@ -22,10 +22,10 @@
         <groupId>net.openhft</groupId>
         <artifactId>java-parent-pom</artifactId>
         <version>1.1.33</version>
-        <relativePath />
+        <relativePath></relativePath>
     </parent>
     <artifactId>chronicle-core</artifactId>
-    <version>2.23ea14-SNAPSHOT</version>
+    <version>2.23ea20-SNAPSHOT</version>
```
JerryShea, line 28: need to rebase?

pom.xml
```diff
@@ -192,7 +192,7 @@
                         <configuration>
                             <referenceVersion>2.23ea0</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
-                            <binaryCompatibilityPercentageRequired>100.0</binaryCompatibilityPercentageRequired>
+                            <!--                            <binaryCompatibilityPercentageRequired>99</binaryCompatibilityPercentageRequired>-->
```
JerryShea, line 195: need to uncomment

src/main/java/net/openhft/chronicle/core/io/AbstractReferenceCounted.java
```diff
@@ -174,21 +176,32 @@ public boolean reservedBy(ReferenceOwner owner) throws IllegalStateException {
...
+            throw new IllegalStateException(message, usedByThreadHere);
         }
         return true;
     }
 
+    public void singleThreadedCheckReset() {
+        usedByThread = null;
+    }
+
+    @Deprecated(/* to be removed in x.25 */)
```
JerryShea, line 207: should add to the comment that instead should use `singleThreadedCheckReset`

src/main/java/net/openhft/chronicle/core/io/AbstractReferenceCounted.java
```diff
@@ -174,21 +176,32 @@ public boolean reservedBy(ReferenceOwner owner) throws IllegalStateException {
...
         if (usedByThread == null && !isUsed)
             return true;
 
         Thread currentThread = Thread.currentThread();
         if (usedByThread == null || !usedByThread.isAlive()) {
             usedByThread = currentThread;
+            usedByThreadHere = new StackTrace("Used here");
         } else if (usedByThread != currentThread) {
-            throw new IllegalStateException(getClass().getName() + " component which is not thread safes used by " + usedByThread + " and " + currentThread);
+            final String message = getClass().getName() + " component which is not thread safes used by " + usedByThread + " and " + currentThread;
```
JerryShea: safes -> safe

JerryShea (approved):
approved but please take a look at the comments!

minborg:
Take a look at the `ThreadConfinementAsserter` in Core.

pom.xml
peter-lawrey: This is to ensure I am using the version I think I am, as this is a significant change that took a while to get right.

pom.xml
peter-lawrey: why?

peter-lawrey:
@minborg Two things we also needed was a stack trace of where it was used in the original thread so we can see where the two accesses are. Also, some components are initialised in one thread and used in another, so we need a means of resetting the state to say the previous thread will no longer touch it.

minborg:
> @minborg Two things we also needed was a stack trace of where it was used in the original thread so we can see where the two accesses are. Also, some components are initialised in one thread and used in another, so we need a means of resetting the state to say the previous thread will no longer touch it.

Yes. This is a classical problem of create vs. use and threads.

Take a look at the core branch `threadsafety` I made which addresses precisely this.

RobAustin (approved):

Issue #412, opened 2022-06-13
Added support for annotation aliases to make them simpler


peter-lawrey:
E.g.
```java
@LongConvertion(Base85LongConverter.class)
@interface Base85 {
}

@Base85 long symbol;

public void symbol(@Base85 long symbol) {
   this.symbol = symbol;
}
```

hft-team-city:
Released in [Chronicle-Core-2.23ea20](https://github.com/OpenHFT/Chronicle-Core/releases/chronicle-core-2.23ea20), [BOM-2.23ea72](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23ea72)

PR #413, opened 2022-06-13
Added support for annotation aliases to make them simpler #412

peter-lawrey:


src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -1174,7 +1171,7 @@ public static Integer getInteger(final String systemPropertyKey, final Integer d
      * <p>
      * This provides a more permissive boolean System systemPropertyKey flag where
      * {@code -Dflag} {@code -Dflag=true} {@code -Dflag=yes} are all accepted.
-     *
+     * <p>
```
RobAustin, line 1174: I think we should remove the <p>'s

src/test/java/net/openhft/chronicle/core/JvmTest.java
```diff
@@ -303,4 +307,25 @@ public void removingTag() {
         assertEquals("value", actual);
     }
 
+    @Test
```
RobAustin, line 310: It would be easier to understand if this was based on a real examples such as Base64 rather than Foo, RealAnno and Anno alias. as this is harder to understand what is going on.

RobAustin (approved):

src/test/java/net/openhft/chronicle/core/JvmTest.java
peter-lawrey: +1

src/main/java/net/openhft/chronicle/core/Jvm.java
peter-lawrey: I can't see 's

PR #420, opened 2022-06-22
Fix #402

### LINKED ISSUE #402: Improve doco of system properties
Re: https://github.com/OpenHFT/Chronicle-Core/#system-properties-from-file

This doco is out of date - there is an alternative to calling Jvm.init() in a static block - we can use Jvm.getInteger etc. as added by @alamar for #354

It would be also useful to refer to the javadoc for Chronicle's opinionated boolean system properties - `net.openhft.chronicle.core.Jvm#getBoolean(java.lang.String)`

j4sm1ne96:
@alamar I have added you in too as you added in the code below and hence may have advice regarding this readMe: 

https://github.com/OpenHFT/Chronicle-Core/blob/ea/src/main/java/net/openhft/chronicle/core/ChronicleInit.java

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -39,7 +39,7 @@ if it can find this file in the current directory. If it can't find this file in
 look in `..` too.
 This is a useful feature if you want to de-clutter your command line.
 
-Chronicle Core class `Jvm` loads the file `system.properties` into the System's properties. To ensure it is loaded early enough, add the following code to your Main
+Chronicle Core class `Jvm` loads the file `system.properties` into the System's properties. To ensure it is loaded early enough, you can load in a static block
 
 [source,java]
 ----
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -40,19 +40,8 @@ look in `..` too.
 This is a useful feature if you want to de-clutter your command line.
 
 Chronicle Core class `Jvm` loads the file `system.properties` into the System's properties.
-To ensure it is loaded early enough, add the following code to your `Main`
```
alamar, line 43: Why did you remove this section? I think it's still relevant. And the next one too. I suggest asking @peter-lawrey

rogersimmons (approved):

README.adoc
```diff
@@ -39,7 +39,16 @@ if it can find this file in the current directory. If it can't find this file in
 look in `..` too.
 This is a useful feature if you want to de-clutter your command line.
 
-Chronicle Core class `Jvm` loads the file `system.properties` into the System's properties.
+Chronicle Core class `Jvm` loads the file `system.properties` into the System's properties. To ensure it is loaded early enough, add the following code to your Main
```
JerryShea: "add the following code to your Main" -> "you can load in a static block"

Also please mention how Jvm.getInteger etc work and how they ensure that system.properties are loaded

PR #426, opened 2022-06-23
consistent thread safety property names

JerryShea:


nicktindall (approved):

src/main/java/net/openhft/chronicle/core/io/SingleThreadedChecked.java
```diff
@@ -1,6 +1,13 @@
 package net.openhft.chronicle.core.io;
 
+import net.openhft.chronicle.core.Jvm;
+
 public interface SingleThreadedChecked {
+    // TODO: remove disable.thread.safety property in x.25
+    boolean DISABLE_SINGLE_THREADED_CHECK =
+            Jvm.getBoolean("disable.single.threaded.check",
+                    Jvm.getBoolean("disable.thread.safety", false));
```
nicktindall, line 9: I still wonder if it'd be better to provide this via a static method call, that would add a layer of indirection to encapsulate how it's determined, leaving the door open to making it more dynamic in the future rather than consumers accessing the field directly.
That seems to be the pattern for all the static fields in `Jvm`, they all seem to be accessed via static method calls.

nicktindall: Dynamism might be useful for e.g. overriding in tests

JerryShea: Why don't we merge as-is, as is this is the smallest possible change

PR #443, opened 2022-08-31
Add ReferenceChangeListener

nicktindall:
This is to support the implementation of reliable expiry in `ReferenceCountedCache`. We should be able to add a `ReferenceCountedListener` which can schedule a cache flush when the refCount goes to 1 for a tracked `ReferenceCounted`.

**Note**
I made it so the `ReferenceCountedListeners` DO NOT get called for releases that happen as part of `warnAndReleaseIfNotReleased`, because I think this method may be called in a finalizer, so we probably don't want to be executing arbitrary code there. This is explained in the javadoc for `warnAndReleaseIfNotReleased`

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/io/DualReferenceCounted.java
+++ b/src/main/java/net/openhft/chronicle/core/io/DualReferenceCounted.java
@@ -50,6 +50,16 @@ public class DualReferenceCounted implements MonitorReferenceCounted {
         return a.createdHere();
     }
 
+    @Override
+    public void addReferenceChangeListener(ReferenceChangeListener referenceChangeListener) {
+        a.addReferenceChangeListener(referenceChangeListener);
+    }
+
+    @Override
+    public void removeReferenceChangeListener(ReferenceChangeListener referenceChangeListener) {
+        a.removeReferenceChangeListener(referenceChangeListener);
+    }
+
     @Deprecated(/* To be removed in x.25 */)
     @Override
     public boolean reservedBy(ReferenceOwner owner) throws IllegalStateException {
--- a/src/main/java/net/openhft/chronicle/core/io/ReferenceChangeListenerManager.java
+++ b/src/main/java/net/openhft/chronicle/core/io/ReferenceChangeListenerManager.java
@@ -1,0 +1,61 @@
+package net.openhft.chronicle.core.io;
+
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+class ReferenceChangeListenerManager {
+
+    private final List<ReferenceChangeListener> referenceChangeListeners;
+    private final ReferenceCounted owner;
+
+    public ReferenceChangeListenerManager(ReferenceCounted owner) {
+        this.owner = owner;
+        referenceChangeListeners = new CopyOnWriteArrayList<>();
+    }
+
+    void add(ReferenceChangeListener referenceChangeListener) {
+        referenceChangeListeners.add(referenceChangeListener);
+    }
+
+    void remove(ReferenceChangeListener referenceChangeListener) {
+        referenceChangeListeners.remove(referenceChangeListener);
+    }
+
+    void notifyAdded(ReferenceOwner referenceOwner) {
+        this.callReferenceChangeListeners(
+                (listener, referenceCounted, lhs, rhs) -> listener.onReferenceAdded(referenceCounted, lhs),
+                referenceOwner, null);
+    }
+
+    void notifyRemoved(ReferenceOwner referenceOwner) {
+        this.callReferenceChangeListeners(
+                (listener, referenceCounted, lhs, rhs) -> listener.onReferenceRemoved(referenceCounted, lhs),
+                referenceOwner, null);
+    }
+
+    void notifyTransferred(ReferenceOwner from, ReferenceOwner to) {
+        this.callReferenceChangeListeners(ReferenceChangeListener::onReferenceTransferred, from, to);
+    }
+
+    private void callReferenceChangeListeners(ListenerInvoker listenerInvoker, ReferenceOwner lhs, ReferenceOwner rhs) {
+        //noinspection ForLoopReplaceableByForEach
+        for (int i = 0; i < referenceChangeListeners.size(); i++) {
+            ReferenceChangeListener listener;
+            try {
+                listener = referenceChangeListeners.get(i);
+            } catch (ArrayIndexOutOfBoundsException e) {
+                // This can happen if a listener is removed concurrently during the iteration
+                continue;
+            }
+            listenerInvoker.invokeListener(listener, owner, lhs, rhs);
+        }
+    }
+
+    public void clear() {
+        referenceChangeListeners.clear();
+    }
+
+    private interface ListenerInvoker {
+        void invokeListener(ReferenceChangeListener listener, ReferenceCounted referenceCounted, ReferenceOwner lhs, ReferenceOwner rhs);
+    }
+}
--- a/src/main/java/net/openhft/chronicle/core/io/ReferenceChangeListenerManager.java
+++ b/src/main/java/net/openhft/chronicle/core/io/ReferenceChangeListenerManager.java
@@ -1,5 +1,7 @@
 package net.openhft.chronicle.core.io;
 
+import org.jetbrains.annotations.Nullable;
+
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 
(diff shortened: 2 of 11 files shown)


src/main/java/net/openhft/chronicle/core/io/DualReferenceCounted.java
```diff
@@ -49,6 +49,18 @@ public StackTrace createdHere() {
...
+    @Override
+    public void addReferenceChangeListener(ReferenceChangeListener referenceChangeListener) {
+        a.addReferenceChangeListener(referenceChangeListener);
+        b.addReferenceChangeListener(referenceChangeListener);
+    }
+
+    @Override
+    public void removeReferenceChangeListener(ReferenceChangeListener referenceChangeListener) {
+        a.removeReferenceChangeListener(referenceChangeListener);
+        b.removeReferenceChangeListener(referenceChangeListener);
```
nicktindall: This is nasty, if one day we had two `ReferenceCounted`'s that both supported listeners. I'm not sure what to do here. The pattern thus far seems to be do things to `a` that make sense to only happen once, but in this case, currently, `a` is the `TracingReferenceCounted`. So I'd have to re-implement the listener handling in there and you'd get a different implementation when reference tracing was turned on or off.
Perhaps `DualReferenceCounted` should have explicitly typed `a` and `b` so we can bake assumptions in. I don't think being generic is very helpful here. Perhaps it should also be internal?

src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
```diff
@@ -55,6 +55,16 @@ public StackTrace createdHere() {
...
 
+    @Override
+    public void addReferenceChangeListener(ReferenceChangeListener referenceChangeListener) {
+        // This is taken care of by VanillaReferenceCounted
+    }
+
+    @Override
+    public void removeReferenceChangeListener(ReferenceChangeListener referenceChangeListener) {
+        // This is taken care of by VanillaReferenceCounted
+    }
```
nicktindall, line 66: This is how I got around `DualReferenceCounted`. I'm not happy with it but also not sure what the best solution is here.

src/main/java/net/openhft/chronicle/core/io/VanillaReferenceCounted.java
```diff
@@ -120,6 +130,7 @@ public void callOnRelease() throws ClosedIllegalStateException {
         releasedHere = Jvm.isResourceTracing() ? new StackTrace("Released here") : null;
         released = true;
         onRelease.run();
+        referenceChangeListeners.clear();
```
nicktindall, line 133: clear the listeners after `releaseLast(...)` has happened.

src/main/java/net/openhft/chronicle/core/io/VanillaReferenceCounted.java
```diff
@@ -175,6 +187,28 @@ public boolean unmonitored() {
...
+
+    private void callReferenceChangeListeners(ChangeType changeType, ReferenceOwner referenceOwner) {
+        //noinspection ForLoopReplaceableByForEach
+        for (int i = 0; i < referenceChangeListeners.size(); i++) {
+            ReferenceChangeListener listener;
+            try {
+                listener = referenceChangeListeners.get(i);
+            } catch (ArrayIndexOutOfBoundsException e) {
+                // This can happen if a listener is removed concurrently during the iteration
+                continue;
```
nicktindall, line 206: I think this can happen, CopyOnWriteArrayList would guarantee this wasn't necessary if we took an iterator but because we're using a classic for-loop I believe we could still be `< referenceChangeListeners.size()` three lines ago and get an OOB exception when we go to get it.

src/main/java/net/openhft/chronicle/core/io/VanillaReferenceCounted.java
```diff
@@ -175,8 +186,49 @@ public boolean unmonitored() {
...
+        }
+    }
+
     private ClosedIllegalStateException newReleasedClosedIllegalStateException() {
         return new ClosedIllegalStateException(type.getName() + " released");
     }
 
+    private interface ListenerInvoker {
+        void invokeListener(ReferenceChangeListener listener, ReferenceCounted referenceCounted, ReferenceOwner lhs, ReferenceOwner rhs);
+    }
```
nicktindall, line 233: Having this interface means we can create non-capturing lambdas in `notifyListenersXXX` so we can re-use `callReferenceChangeListeners` for adds, removes and transfers

src/main/java/net/openhft/chronicle/core/io/DualReferenceCounted.java
nicktindall: DualReferenceCounted is now deprecated

src/main/java/net/openhft/chronicle/core/io/ReferenceChangeListenerManager.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private void callReferenceChangeListeners(ListenerInvoker listenerInvoker, ReferenceOwner lhs, ReferenceOwner rhs) {
+        //noinspection ForLoopReplaceableByForEach
+        for (int i = 0; i < referenceChangeListeners.size(); i++) {
+            ReferenceChangeListener listener;
+            try {
+                listener = referenceChangeListeners.get(i);
+            } catch (ArrayIndexOutOfBoundsException e) {
+                // This can happen if a listener is removed concurrently during the iteration
+                continue;
```
nicktindall, line 50: I think this can happen, `CopyOnWriteArrayList` would guarantee this wasn't necessary if we took an iterator but because we're using a classic for-loop I believe we could still be `< referenceChangeListeners.size()` three lines ago and get an OOB exception when we go to get it.

src/main/java/net/openhft/chronicle/core/io/ReferenceChangeListenerManager.java
```diff
@@ -0,0 +1,63 @@
...
+
+    void notifyAdded(ReferenceOwner referenceOwner) {
+        this.callReferenceChangeListeners(
+                (listener, referenceCounted, lhs, rhs) -> listener.onReferenceAdded(referenceCounted, lhs),
+                referenceOwner, null);
+    }
+
+    void notifyRemoved(@Nullable ReferenceOwner referenceOwner) {
+        this.callReferenceChangeListeners(
+                (listener, referenceCounted, lhs, rhs) -> listener.onReferenceRemoved(referenceCounted, lhs),
```
nicktindall, line 34: these are non-capturing lambdas

src/main/java/net/openhft/chronicle/core/io/ReferenceCountedTracer.java
```diff
@@ -38,6 +38,13 @@ default void throwExceptionIfReleased() throws ClosedIllegalStateException {
             throw new ClosedIllegalStateException("Released");
     }
 
+    /**
+     * Release any remaining references, and log a warning if there were any references to release.
+     * <p>
+     * Intended to be called by a finalizer or in a test, to confirm that references are being released correctly
+     * <p>
+     * Note: This will not cause any {@link ReferenceChangeListener}s to fire as it's really just a sanity check
```
nicktindall, line 46: :point_up:

src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
```diff
@@ -85,40 +97,32 @@ private boolean tryReserve(ReferenceOwner id, boolean must) throws IllegalStateE
             }
             StackTrace stackTrace = references.get(id);
             if (stackTrace == null)
-                references.putIfAbsent(id, stackTrace("reserve", id));
+                references.put(id, stackTrace("reserve", id));
```
nicktindall, line 100: Not sure why this needed to be `putIfAbsent` given we're in a `synchronized(references)` block and we establish that there's no entry on the previous line

src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
```diff
@@ -127,6 +132,37 @@ public void release(ReferenceOwner id) throws IllegalStateException {
...
+            }
+            if (references.remove(from) == null) {
+                throw throwInvalidReleaseException(from);
+            }
+            releases.put(from, stackTrace("reserveTransfer", from));
+            references.put(to, stackTrace("reserveTransfer", to));
+            releases.remove(to);
+        }
+        referenceChangeListeners.notifyTransferred(from, to);
+    }
```
nicktindall, line 150: Had to add this because it was using the default of `reserve(to);release(from)` previously, but we want to fire the specific event of a transfer

src/main/java/net/openhft/chronicle/core/io/VanillaReferenceCounted.java
```diff
@@ -118,12 +124,13 @@ public void release(ReferenceOwner id) throws ClosedIllegalStateException {
         }
     }
 
-    public void callOnRelease() throws ClosedIllegalStateException {
+    private void callOnRelease() throws ClosedIllegalStateException {
```
nicktindall, line 127: This was only used by VanillaReferenceCounted, it seems like an internal method to me

JerryShea (approved):

peter-lawrey (approved):

PR #446, opened 2022-09-02
Deprecate DualReferenceCounted

### LINKED ISSUE #445: DualReferenceCounted introduces risk of deadlock
`DualReferenceCounted` wraps an instance of `VanillaReferenceCounted` and one of `TracingReferenceCounted`, delegating all calls that change state (e.g. `reserve`/`release`/`releaseLast`) to both of them. These calls are made in synchronized blocks so as to ensure both child `ReferenceCounted` objects are called in sequence atomically.

This presents a number of issues, 
- If one of the calls fails, the two `ReferenceCounted`s can get out of sync (this currently happens if you called `releaseLast`, with a reference that isn't the last)
- The `onRelease` handler in the children are called in the synchronized block, [introducing the possibility of a deadlock](https://github.com/OpenHFT/Chronicle-Core/blob/e243bfa5095657d6e3ecc877a6302bec41f2c13e/src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java#L121)

[The addition of the ReferenceChangeHandler](https://github.com/OpenHFT/Chronicle-Core/pull/443) adds to this risk. I propose we deprecate `DualReferenceHandler` and use either `TracingReferenceHandler` or `VanillaReferenceHandler` individually depending on if tracing is turned on.

nicktindall:
Fixes #445

I assume the reason for DualReferenceCounted was as a kind of integration test to confirm that the behaviour of Tracing and Vanilla `ReferenceCounted`s was consistent (it wasn't, see changes to `TracingReferenceCounted#releaseLast`).

I think a better way to ensure their consistency is to implement contract tests for the interfaces and run all the implementations through them.

I also found some inconsistencies between the contract and the implementations.

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/Memory.java
+++ b/src/main/java/net/openhft/chronicle/core/Memory.java
@@ -164,6 +164,10 @@ public interface Memory {
 
     boolean compareAndSwapLong(Object object, long offset, long expected, long value) throws MisAlignedAssertionError;
 
+    int getAndSetInt(long address, int value) throws MisAlignedAssertionError;
+
+    int getAndSetInt(Object object, long offset, int value) throws MisAlignedAssertionError;
+
     int pageSize();
 
     byte readVolatileByte(long address);
--- a/src/main/java/net/openhft/chronicle/core/io/AbstractCloseable.java
+++ b/src/main/java/net/openhft/chronicle/core/io/AbstractCloseable.java
@@ -43,7 +43,9 @@ public abstract class AbstractCloseable implements ReferenceOwner, ManagedClosea
     @Deprecated(/* remove in x.25 */)
     protected static final boolean DISABLE_THREAD_SAFETY = DISABLE_SINGLE_THREADED_CHECK;
     protected static final boolean DISABLE_DISCARD_WARNING = Jvm.getBoolean("disable.discard.warning", false);
-    protected static final boolean STRICT_DISCARD_WARNING = Jvm.getBoolean("strict.discard.warning", false);
+    @SuppressWarnings("DeprecatedIsStillUsed")
+    @Deprecated(/* remove in x.25 */)
+    protected static final boolean STRICT_DISCARD_WARNING;
 
     protected static final long WARN_NS = (long) (Jvm.getDouble("closeable.warn.secs", 0.02) * 1e9);
 
@@ -57,6 +59,10 @@ public abstract class AbstractCloseable implements ReferenceOwner, ManagedClosea
         if (Jvm.isResourceTracing())
             enableCloseableTracing();
         CLOSED_OFFSET = UnsafeMemory.unsafeObjectFieldOffset(Jvm.getField(AbstractCloseable.class, "closed"));
+        if (Jvm.getProperty("strict.discard.warning") != null) {
+            Jvm.warn().on(AbstractCloseable.class, "strict.discard.warning is deprecated and has no effect, it will be removed in x.25");
+        }
+        STRICT_DISCARD_WARNING = Jvm.getBoolean("strict.discard.warning", false);
     }
 
     private final transient StackTrace createdHere;
--- a/src/main/java/net/openhft/chronicle/core/io/ReferenceCountedTracer.java
+++ b/src/main/java/net/openhft/chronicle/core/io/ReferenceCountedTracer.java
@@ -28,10 +28,7 @@ public interface ReferenceCountedTracer extends ReferenceCounted {
     @NotNull
     static ReferenceCountedTracer onReleased(final Runnable onRelease, Supplier<String> uniqueId, Class<?> type) {
         return Jvm.isResourceTracing()
-                ? new DualReferenceCounted(
-                new TracingReferenceCounted(onRelease, uniqueId.get(), type),
-                new VanillaReferenceCounted(() -> {
-                }, type))
+                ? new TracingReferenceCounted(onRelease, uniqueId.get(), type)
                 : new VanillaReferenceCounted(onRelease, type);
     }
 
--- a/src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
+++ b/src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
@@ -6,8 +6,6 @@ package net.openhft.chronicle.core.io;
 
 import net.openhft.chronicle.core.Jvm;
 import net.openhft.chronicle.core.StackTrace;
-import net.openhft.chronicle.core.onoes.ExceptionHandler;
-import net.openhft.chronicle.core.onoes.Slf4jExceptionHandler;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.Collections;
@@ -42,8 +40,8 @@ public final class TracingReferenceCounted implements MonitorReferenceCounted {
         if (id instanceof ReferenceCounted)
             s += " refCount=" + ((ReferenceCounted) id).refCount();
         try {
-            if (id instanceof Closeable)
-                s += " closed=" + ((Closeable) id).isClosed();
+            if (id instanceof QueryCloseable)
+                s += " closed=" + ((QueryCloseable) id).isClosed();
         } catch (NullPointerException ignored) {
             // not initialised
         }
@@ -55,6 +53,7 @@ public final class TracingReferenceCounted implements MonitorReferenceCounted {
         return createdHere;
     }
 
+    @Deprecated
     @Override
     public boolean reservedBy(ReferenceOwner owner) throws IllegalStateException {
         if (references.containsKey(owner))
@@ -139,15 +138,13 @@ public final class TracingReferenceCounted implements MonitorReferenceCounted {
 
     @Override
     public void releaseLast(ReferenceOwner id) throws IllegalStateException {
-        if (references.size() <= 1) {
+        Exception e0 = null;
+        try {
             release(id);
-        } else {
-            Exception e0 = null;
-            try {
-                release(id);
-            } catch (Exception e) {
-                e0 = e;
-            }
+        } catch (Exception e) {
+            e0 = e;
+        }
+        if (references.size() > 0) {
             IllegalStateException ise = new IllegalStateException(type.getName() + " still reserved " + referencesAsString(), createdHere);
             synchronized (references) {
                 references.values().forEach(ise::addSuppressed);
@@ -156,6 +153,8 @@ public final class TracingReferenceCounted implements MonitorReferenceCounted {
                 ise.addSuppressed(e0);
             throw ise;
         }
+        if (e0 != null)
+            Jvm.rethrow(e0);
     }
 
     @Override
@@ -230,11 +229,18 @@ public final class TracingReferenceCounted implements MonitorReferenceCounted {
 
     @Override
     public void warnAndReleaseIfNotReleased() {
-        if (refCount() > 0) {
-            if (!unmonitored && !AbstractCloseable.DISABLE_DISCARD_WARNING) {
-                ExceptionHandler warn = AbstractCloseable.STRICT_DISCARD_WARNING ? Jvm.warn() : Slf4jExceptionHandler.WARN;
-                warn.on(type, "Discarded without being released by " + referencesAsString(), createdHere);
+        boolean runOnRelease = false;
+        synchronized (references) {
+            if (refCount() > 0) {
+                if (!unmonitored && !AbstractCloseable.DISABLE_DISCARD_WARNING) {
+                    Jvm.warn().on(type, "Discarded without being released by " + referencesAsString(), createdHere);
+                }
+                references.clear();
+                runOnRelease = true;
             }
+        }
+        // Run outside the synchronized block to avoid risk of deadlock
+        if (runOnRelease) {
             onRelease.run();
         }
     }
--- a/src/main/java/net/openhft/chronicle/core/io/VanillaReferenceCounted.java
+++ b/src/main/java/net/openhft/chronicle/core/io/VanillaReferenceCounted.java
@@ -22,7 +22,6 @@ import net.openhft.chronicle.core.Jvm;
 import net.openhft.chronicle.core.StackTrace;
 import net.openhft.chronicle.core.UnsafeMemory;
 import net.openhft.chronicle.core.annotation.UsedViaReflection;
-import net.openhft.chronicle.core.onoes.Slf4jExceptionHandler;
 
 import static net.openhft.chronicle.core.io.TracingReferenceCounted.asString;
 
@@ -53,6 +52,7 @@ public final class VanillaReferenceCounted implements MonitorReferenceCounted {
         return null;
     }
 
+    @Deprecated
     @Override
     public boolean reservedBy(ReferenceOwner owner) throws IllegalStateException {
         if (refCount() <= 0)
@@ -97,6 +97,10 @@ public final class VanillaReferenceCounted implements MonitorReferenceCounted {
         return UnsafeMemory.INSTANCE.compareAndSwapInt(this, VALUE, from, to);
     }
 
+    private int valueGetAndSet(int to) {
+        return UnsafeMemory.INSTANCE.getAndSetInt(this, VALUE, to);
+    }
+
     @Override
     public void release(ReferenceOwner id) throws ClosedIllegalStateException {
         for (; ; ) {
@@ -124,20 +128,21 @@ public final class VanillaReferenceCounted implements MonitorReferenceCounted {
 
     @Override
     public void releaseLast(ReferenceOwner id) throws IllegalStateException {
-        for (; ; ) {
-            int v = value;
-            if (v <= 0) {
-                if (Jvm.supportThread())
-                    break;
-                throw newReleasedClosedIllegalStateException();
+        Exception thrownException = null;
+        try {
+            release(id);
+        } catch (Exception e) {
+            thrownException = e;
+        }
+        if (value > 0) {
+            final IllegalStateException ise = new IllegalStateException(type.getName() + " not the last released");
+            if (thrownException != null) {
+                ise.addSuppressed(thrownException);
             }
-            if (v > 1) {
-                throw new IllegalStateException(type.getName() + " not the last released");
-            }
-            if (valueCompareAndSet(1, 0)) {
-                callOnRelease();
-                break;
-            }
+            throw ise;
+        }
+        if (thrownException != null) {
+            Jvm.rethrow(thrownException);
         }
     }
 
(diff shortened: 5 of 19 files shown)


src/main/java/net/openhft/chronicle/core/io/AbstractCloseable.java
```diff
@@ -43,7 +43,9 @@ public abstract class AbstractCloseable implements ReferenceOwner, ManagedClosea
     @Deprecated(/* remove in x.25 */)
     protected static final boolean DISABLE_THREAD_SAFETY = DISABLE_SINGLE_THREADED_CHECK;
     protected static final boolean DISABLE_DISCARD_WARNING = Jvm.getBoolean("disable.discard.warning", false);
-    protected static final boolean STRICT_DISCARD_WARNING = Jvm.getBoolean("strict.discard.warning", false);
+    @SuppressWarnings("DeprecatedIsStillUsed")
+    @Deprecated(/* remove in x.25 */)
+    protected static final boolean STRICT_DISCARD_WARNING;
```
nicktindall, line 48: I don't understand the purpose of this, when true it seemed to use `Jvm.warn` instead of `Slf4jExceptionHandler.WARN` to log the warning that something was discarded without being closed. Using the latter made it hard to test that we warned when that happened, so I just removed this. Happy to un-remove if there's a good reason for its existence.

src/main/java/net/openhft/chronicle/core/io/ReferenceCountedTracer.java
```diff
@@ -28,10 +28,7 @@ public interface ReferenceCountedTracer extends ReferenceCounted {
     @NotNull
     static ReferenceCountedTracer onReleased(final Runnable onRelease, Supplier<String> uniqueId, Class<?> type) {
         return Jvm.isResourceTracing()
-                ? new DualReferenceCounted(
-                new TracingReferenceCounted(onRelease, uniqueId.get(), type),
-                new VanillaReferenceCounted(() -> {
-                }, type))
+                ? new TracingReferenceCounted(onRelease, uniqueId.get(), type)
```
nicktindall, line 31: Just use `TracingReferenceCounted` instead of dual

src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
```diff
@@ -142,18 +140,10 @@ public void releaseLast(ReferenceOwner id) throws IllegalStateException {
         if (references.size() <= 1) {
             release(id);
         } else {
-            Exception e0 = null;
-            try {
-                release(id);
-            } catch (Exception e) {
-                e0 = e;
-            }
```
nicktindall, line 150: This behaviour I think is incorrect, it would perform the release regardless of whether the reference was the last, then throw the exception if it wasn't the last, this differed from the behaviour in `VanillaReferenceCounted` where the count was only decremented if it was <= 1 (which feels more correct)

Happy to change Vanilla to match this behaviour instead if we think it is correct.

Whatever happens they should probably be consistent. This was where I noticed the problem with Dual.

src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
```diff
@@ -230,11 +220,18 @@ public void throwExceptionIfReleased() throws ClosedIllegalStateException {
...
+            if (refCount() > 0) {
+                if (!unmonitored && !AbstractCloseable.DISABLE_DISCARD_WARNING) {
+                    Jvm.warn().on(type, "Discarded without being released by " + referencesAsString(), createdHere);
+                }
+                references.clear();
+                runOnRelease = true;
             }
+        }
+        // Run outside the synchronized block to avoid risk of deadlock
+        if (runOnRelease) {
```
nicktindall, line 243: I changed warnAndRelease to actually clear the references, so any subsequent calls to reserve or release would fail. I did this with Tracing and Vanilla

src/test/java/net/openhft/chronicle/core/io/TracingReferenceCountedTest.java
```diff
@@ -0,0 +1,78 @@
...
+        ReferenceOwner a = ReferenceOwner.temporary("a");
+        assertThrows(IllegalStateException.class, () -> referenceCounted.release(a));
+    }
+
+    /**
+     * This is actually a deviation from the contract, it should return false rather than throwing
+     * <p>
+     * Should we fix the contract or fix the behaviour?
+     */
+    @Test
```
nicktindall, line 45: Tracing and Vanilla both throw `IllegalStateExceptions` when `reservedBy` is called with an owner that doesn't have a reservation. The interface says they should return false, we should change the implementations or the interface javadoc to be consistent. I've changed neither so far.

src/main/java/net/openhft/chronicle/core/io/ReferenceCountedTracer.java
peter-lawrey: As long as we are confident the tests are comprehensive enough, this is fine. 
What is the benefit of this change?

src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
peter-lawrey: When you check which reference still holds on to the resource,  you would have two stack traces of owners not released when one of them was actually attempted. i.e. isn't the source of an issue. This is likely to be more confusing than useful.

src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
peter-lawrey: +1 The method name does imply that they should be released.

src/main/java/net/openhft/chronicle/core/io/VanillaReferenceCounted.java
```diff
@@ -158,10 +157,17 @@ public void throwExceptionIfNotReleased() throws IllegalStateException {
...
+            final int refCount = refCount();
+            if (refCount == 0) {
+                break;
+            }
+            if (valueCompareAndSet(refCount, 0)) {
+                if (!unmonitored && !AbstractCloseable.DISABLE_DISCARD_WARNING)
+                    Jvm.warn().on(type, "Discarded without being released");
+                callOnRelease();
+                break;
+            }
```
peter-lawrey: This extra code reduces the chance of showing a warning when you want it to detect incorrect behaviour as early as possible.

peter-lawrey: If anything, you want to more reliably show a warning than avoid it.

src/test/java/net/openhft/chronicle/core/io/TracingReferenceCountedTest.java
peter-lawrey: The interface should show the contract that all implementations follow. The Vanilla can't detect duplicate reserves so it can't throw an ISE. The contract could say it may or may not throw an ISE depending on settings.

peter-lawrey (changes requested):
Significant improvements, and highlighted the need to document some unclear behaviour.

src/main/java/net/openhft/chronicle/core/io/ReferenceCountedTracer.java
nicktindall: See https://github.com/OpenHFT/Chronicle-Core/issues/445 (linked in the description & commit). DualReferenceCounted means event handlers e.g. the current `onRelease` handler and the proposed `ReferenceChangeListener` will be executed inside a synchronized block. We shouldn't execute arbitrary code while holding a resource as it introduces the risk of a deadlock. The first use of `ReferenceChangeListener`, triggering a cache flush in `ReferenceCountedCache`, did deadlock when executed synchronously.

src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
nicktindall: Good point, I'll change Vanilla & the contract test to match this behaviour instead

src/test/java/net/openhft/chronicle/core/io/TracingReferenceCountedTest.java
nicktindall: The interface states that the two possible return values are `true`/`false` (as you'd expect given it returns a boolean). There is no mention of any exceptional cases. It accounts for the fact that some implementations may not know by stating

> Returns <code>true</code> if not sure.

(see below)
```
    /**
     * Best effort check the owner has reserved it. Returns <code>true</code> if not sure.
     *
     * @param owner to check
     * @return false if the owner definitely doesn't own it.
     */
```

I think we should never throw an exception, otherwise if someone wants to know if something is reserved by or not they need to implement
```
boolean actuallyReservedBy() {
    try {
        return bytes.reservedBy(owner);
    } catch (IllegalStateException e) {
        // assume that exception means definitely not, as it seems to
        return false;
    }
}
```
Which seems pretty clunky. If you look at the current implementations, the return value is redundant. None of them will ever return false.

I checked where it's used (see [here](https://github.com/search?p=2&q=org%3AOpenHFT+reservedBy&type=Code), [here](https://github.com/search?q=org%3AChronicleEnterprise+reservedBy&type=Code) and [here](https://github.com/search?q=org%3AChronicleSoftware+reservedBy&type=Code)) and it's only used in assertions, if we return false instead of throwing an exception, we just get an `AssertionError` instead of and `IllegalStateException` when assertions are turned on.

What do you think?

src/main/java/net/openhft/chronicle/core/io/VanillaReferenceCounted.java
nicktindall: Good point, I'll change that.

src/main/java/net/openhft/chronicle/core/io/VanillaReferenceCounted.java
```diff
@@ -124,20 +123,21 @@ public void callOnRelease() throws ClosedIllegalStateException {
...
+        }
+        if (value > 0) {
+            final IllegalStateException ise = new IllegalStateException(type.getName() + " not the last released");
+            if (thrownException != null) {
+                ise.addSuppressed(thrownException);
             }
+            throw ise;
+        }
+        if (thrownException != null) {
+            Jvm.rethrow(thrownException);
```
nicktindall, line 145: As discussed this behaviour was changed to match that of `TracingReferenceCounted`, i.e. when the reference released is not the last, it will be released then an exception will be thrown.

src/main/java/net/openhft/chronicle/core/io/VanillaReferenceCounted.java
```diff
@@ -97,6 +96,10 @@ private boolean valueCompareAndSet(int from, int to) {
         return UnsafeMemory.INSTANCE.compareAndSwapInt(this, VALUE, from, to);
     }
 
+    private int valueGetAndSet(int to) {
+        return UnsafeMemory.INSTANCE.getAndSetInt(this, VALUE, to);
```
nicktindall, line 101: I had to add this atomically get and set the value.

src/main/java/net/openhft/chronicle/core/io/VanillaReferenceCounted.java
```diff
@@ -158,9 +162,9 @@ public void throwExceptionIfNotReleased() throws IllegalStateException {
 
     @Override
     public void warnAndReleaseIfNotReleased() throws ClosedIllegalStateException {
-        if (refCount() > 0) {
+        if (valueGetAndSet(0) > 0) {
             if (!unmonitored && !AbstractCloseable.DISABLE_DISCARD_WARNING)
-                Slf4jExceptionHandler.WARN.on(type, "Discarded without being released");
+                Jvm.warn().on(type, "Discarded without being released");
             callOnRelease();
```
nicktindall, line 169: We do atomic get-and-set to ensure we were the ones to take the value to zero and it's safe to `callOnRelease`

src/test/java/net/openhft/chronicle/core/io/TracingReferenceCountedTest.java
nicktindall: Another option might be to deprecate the current method and change it to `assertReservedBy(owner)`, which could throw an exception for the definitely not case and not return anything. (i.e. current behaviour without the redundant return value)

src/main/java/net/openhft/chronicle/core/Memory.java
```diff
@@ -164,6 +164,10 @@ default long heapUsed() {
 
     boolean compareAndSwapLong(Object object, long offset, long expected, long value) throws MisAlignedAssertionError;
 
+    int getAndSetInt(long address, int value) throws MisAlignedAssertionError;
+
+    int getAndSetInt(Object object, long offset, int value) throws MisAlignedAssertionError;
```
nicktindall, line 169: I needed this for `VanillaReferenceCounted#warnAndReleaseIfNotReleased`. Not sure if we want to call this `getAndSet` (consistent with `Unsafe`) or `readAndWrite` (consistent with `Memory`). `compareAndSwap` is consistent with `Unsafe`, so I went with `getAndSet`

src/test/java/net/openhft/chronicle/core/io/ReferenceCountedContractTest.java
```diff
@@ -0,0 +1,242 @@
...
+    public void reservedByWillReturnTrueWhenOwnerHasReferenceReserved() {
+        ReferenceCounted referenceCounted = createReferenceCounted();
+
+        ReferenceOwner a = ReferenceOwner.temporary("a");
+        referenceCounted.reserve(a);
+        assertTrue(referenceCounted.reservedBy(a));
+    }
+
+    @Test
+    public void implementationsShouldBeThreadSafe() throws InterruptedException {
```
nicktindall, line 164: Added this to weed out any thread-safety issues in ReferenceCounted implementations

src/test/java/net/openhft/chronicle/core/io/TracingReferenceCountedTest.java
nicktindall: I did the change to make them consistent with the current interface (i.e. not throwing exceptions). The tracing reference counted will still log a warning when false to distinguish between never-held and held-then-released.

It went through build-all OK, but happy to do something else if you think there's a better solution.

src/main/java/net/openhft/chronicle/core/io/AbstractCloseable.java
peter-lawrey: The purpose was that Jvm.warn() are checked in tests and but Slf4j..WARN isn't; 
We have another means of ignoring warnings, which could be ignored in the XxxxTestCommon instead.

src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
peter-lawrey: We don't have a test to show the expected message when the resources check fails.

src/test/java/net/openhft/chronicle/core/io/TracingReferenceCountedTest.java
peter-lawrey: It's worth considering that the common case is that this level of tracing is off, and this is only useful when tracing is turned on. It's possible this method is no longer useful.

src/main/java/net/openhft/chronicle/core/io/VanillaReferenceCounted.java
peter-lawrey: What does this tell you that CAS doesn't?

nicktindall: It allows us to atomically set the value and get the value that was before we set it, that way we can know if we were the ones responsible for changing the value from >=  1 to zero and therefore whether it's safe to call `onRelease`. We _could_ implement `getAndSet` with `compareAndSet` in a loop, but `getAndSet` means we don't have to do that.

src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
nicktindall: Sorry do you mean to add a test for the specific messages in the exceptions thrown by `release`? I think if we care that much about specific exceptions being thrown we should make specific exception sub-classes rather than testing message text.

There are tests for the specific messages that are logged when `reservedBy` returns false see [here](https://github.com/OpenHFT/Chronicle-Core/pull/446/files#diff-0149803deb8141c24d380ac84a76e201ca92dfd436d69f44d028ca1012302002R48) and [here](https://github.com/OpenHFT/Chronicle-Core/pull/446/files#diff-0149803deb8141c24d380ac84a76e201ca92dfd436d69f44d028ca1012302002R57)

JerryShea (approved):

src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
nicktindall: I added [some more tests around the specific errors thrown when releaseLast fails](https://github.com/OpenHFT/Chronicle-Core/pull/446/commits/da82b5b5f9312cb053ad993e951c2c807ca557c3)

src/test/java/net/openhft/chronicle/core/io/TracingReferenceCountedTest.java
nicktindall: I think if we just make the implementation match the interface as I've done, I'm happy to leave it there or it's becoming scope creep. There's also [a bunch of logic in TracingReferenceCounted](https://github.com/OpenHFT/Chronicle-Core/blob/b546716e3ef70139567886a3d6873c3167d7edff/src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java#L180) that looks like it could do with a tidy up and some more testing, but again it's scope creep for this PR. I'm happy where it's at for now and if we want to do more I suggest we create issues and more PRs later.

src/main/java/net/openhft/chronicle/core/io/TracingReferenceCounted.java
peter-lawrey: I can't see the commits, but I believe you

src/test/java/net/openhft/chronicle/core/io/TracingReferenceCountedTest.java
peter-lawrey: Can you mark it as deprecated for removal in x.25, and leave it for now ?

src/main/java/net/openhft/chronicle/core/io/VanillaReferenceCounted.java
peter-lawrey: We can't force the release of the resource. If releaseLast is called while another thread still uses the resource, we can't have it SEG faulting.

peter-lawrey (approved):
These changes look like a good improvement

PR #448, opened 2022-09-23
Remove deprecated code in 2.24

alamar:


src/main/java/net/openhft/chronicle/core/io/AbstractReferenceCounted.java
```diff
@@ -185,7 +185,7 @@ public void warnAndReleaseIfNotReleased() throws ClosedIllegalStateException {
         referenceCounted.warnAndReleaseIfNotReleased();
     }
 
-    @Deprecated
+    @Deprecated(/* To be removed in 2.25 */)
```
JerryShea, line 188: please use `x.25` as that is the convention used elsewhere

JerryShea (approved):

src/main/java/net/openhft/chronicle/core/internal/threads/VanillaThreadLock.java
```diff
@@ -1,184 +0,0 @@
-/*
```
nicktindall, line 1: Should we be deleting the ThreadLock interface as well, VanillaThreadLock was the only implementation and it was introduced with VanillaThreadLock in 4c1b33d49f432c18f54a9954802ed8f22d1acb5a

nicktindall (approved):

PR #456, opened 2022-10-15
log a warning with link to doco for command line args for Java >= 11

JerryShea:
As customers do not necessarily know about these command line flags


nicktindall (approved):
🙌

PR #458, opened 2022-10-26
Make tracked exceptions tolerate being concurrently added to and iterated over

nicktindall:
This is an approach that could work

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/Jvm.java
+++ b/src/main/java/net/openhft/chronicle/core/Jvm.java
@@ -46,6 +46,7 @@ import java.nio.file.Paths;
 import java.util.*;
 import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -811,7 +812,7 @@ public static Map<ExceptionKey, Integer> recordExceptions(boolean debug, boolean
     public static Map<ExceptionKey, Integer> recordExceptions(final boolean debug,
                                                               final boolean exceptionsOnly,
                                                               final boolean logToSlf4j) {
-        final Map<ExceptionKey, Integer> map = Collections.synchronizedMap(new LinkedHashMap<>());
+        final Map<ExceptionKey, Integer> map = new ConcurrentSkipListMap<>(Comparator.comparing(ExceptionKey::nanoTimestamp));
```
nicktindall, line 826: If we use a wrapped LinkedHashMap, the iterator returned from entrySet() won't be tolerate concurrent modification, but if we use a truly concurrent map, the iterator will. Unfortunately there's no ordered concurrent map, only sorted concurrent maps, so if we add a nano timestamp we'll get the ordering by insertion time we're after?

src/main/java/net/openhft/chronicle/core/onoes/ExceptionKey.java
```diff
@@ -82,7 +89,8 @@ public String toString() {
         if (throwable != null)
             throwable.printStackTrace(new PrintWriter(sw));
         return "ExceptionKey{" +
-                "level=" + level +
+                "nanoTimestamp=" + nanoTimestamp +
```
nicktindall, line 92: of two minds whether to include this in the toString, and if we do, perhaps we format it?

src/main/java/net/openhft/chronicle/core/Jvm.java
nicktindall: I wonder though if inserting sorted data (the timestamps will be sequential naturally) ends up producing some kind of pathological skip list. Time to dig out the data structures book.

JerryShea (approved):

PR #463, opened 2022-12-05
Return Closeable to allow cancellation of scheduled tasks in Timer…

### LINKED ISSUE #462: Add a way to cancel scheduled handlers submitted to a Timer
Currently there's no way to cancel a task submitted via `Timer#schedule*` methods. ScheduledEventHandler implements closeable and is coded to throw InvalidHandlerException after it's closed but there's no way to get a reference to the handler to do that.

nicktindall:
…add tests for Timer, make Timer's TimeProvider configurable. Fixes #462.

### DIFF
--- a/src/test/java/net/openhft/chronicle/core/threads/TimerTest.java
+++ b/src/test/java/net/openhft/chronicle/core/threads/TimerTest.java
@@ -1,0 +1,131 @@
+package net.openhft.chronicle.core.threads;
+
+import net.openhft.chronicle.core.time.SetTimeProvider;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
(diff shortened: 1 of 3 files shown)


src/test/java/net/openhft/chronicle/core/threads/TimerTest.java
```diff
@@ -0,0 +1,131 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+
+@ExtendWith(MockitoExtension.class)
+class TimerTest {
+
+    public static final int INITIAL_DELAY_MS = 1_000;
+    public static final int PERIOD_MS = 2_000;
+    @Mock
+    private EventLoop eventLoop;
```
nicktindall, line 24: There are no EventLoop implementations in Core so we have no choice but to mock here, also the tests read nicely and involve no threads with the mocked EventLoop.

JerryShea (approved):

PR #471, opened 2023-01-23
Add ChronicleInit option to initialize from ServiceLoader

### LINKED ISSUE #470: Add ServiceLoader-based runnable discovery to ChronicleInit
In addition to property-based initialization it should be possible to use ServiceLoader/META-INF-based initialization.

alamar:
closed #470

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/ChronicleInit.java
+++ b/src/main/java/net/openhft/chronicle/core/ChronicleInit.java
@@ -23,17 +23,23 @@ import java.util.ServiceLoader;
 /**
  * Handles application code which must be loaded first/run and may override system properties.
  * <p>
- * A {@link Runnable} fully qualified class name may be provided via {@code chronicle.init.runnable} property.
+ * A {@link Runnable} fully qualified class name may be provided via {@code chronicle.init.runnable} property,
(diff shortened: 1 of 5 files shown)


src/main/java/net/openhft/chronicle/core/ChronicleInit.java
```diff
@@ -46,6 +52,20 @@ private ChronicleInit() {
...
         }
+
+        try {
+            ServiceLoader<ChronicleInitRunnable> runnableLoader = ServiceLoader.load(ChronicleInitRunnable.class);
+
+            for (Runnable runnable : runnableLoader) {
+                try {
+                    runnable.run();
+                } catch (Exception ex) {
+                    ex.printStackTrace();
```
JerryShea, line 69: Sonar will complain about these

alamar: What's the alternative here? No logging subsystem is expected to be up at this point.

JerryShea: Fair point!

glukos (approved):

PR #477, opened 2023-02-13
Interfaces for Observable prototype (Experimental, do not merge)

nicktindall:
These provide some interfaces that can be implemented to allow system state to be dumped. It abstracts over how the output is written, there is a basic StateReporter that writes to an Appendable (Bytes/StringBuilder/PrintWriter) and there is one [implemented in Wire](https://github.com/OpenHFT/Chronicle-Wire/pull/622) that writes to a Wire.

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/observable/AppendableStateReporter.java
+++ b/src/main/java/net/openhft/chronicle/core/observable/AppendableStateReporter.java
@@ -1,0 +1,90 @@
+package net.openhft.chronicle.core.observable;
+
+import net.openhft.chronicle.core.io.IORuntimeException;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.IdentityHashMap;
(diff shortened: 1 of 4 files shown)


src/main/java/net/openhft/chronicle/core/observable/AppendableStateReporter.java
```diff
@@ -0,0 +1,90 @@
...
+        } else {
+            if (value instanceof Observable) {
+                Observable observable = (Observable) value;
+                writeIndent().append(ID_PROPERTY).append(": ").append(observable.idString()).newLine();
+                if (!seenObjects.containsKey(value)) {
+                    seenObjects.put(value, value);
+                    observable.dumpState(this);
+                }
+            } else {
+                writeIndent().append(value.toString()).newLine();
```
JerryShea, line 51: should put try/catch round this

PR #480, opened 2023-02-21
De-duplicate asDouble method, expose it publicly in Maths.asDouble(...) 

### LINKED ISSUE #479: De-duplicate toDouble implementation and expose it publicly
We have two implementations of toDouble that do approximately the same thing. To avoid more being created we should standardise on one and make it reusable.

nicktindall:
Fixes #479

I added the perf test so the two implementations could be compared. The one from `UnsafeText.asDouble` is much faster than the one from `StringUtils.asDouble`, and other than for negative decimal places, which `StringUtils.asDouble` didn't seem to support, they produce the same values for the inputs I tested.

Checkout 69619f7e4c48852d7fc29172b9e13ab4f1fbce56 to run the JLBH comparison

This is also implemented in [a third place](https://github.com/OpenHFT/Chronicle-Bytes/blob/9a20a7ab9b7fe1c50d26bac0776b07c4f32b4470/src/main/java/net/openhft/chronicle/bytes/internal/BytesInternal.java#L2144), I ran that implementation through the JLBH test and it was marginally slower than the `UnsafeText.asDouble` implementation for the test values. If approved we should remove that too.

```
-------------------------------- SUMMARY (StringUtils.asDouble) us ---------------------------------
Percentile   run1         run2         run3      % Variation
50.0:           13.26        13.23        13.55         1.59
90.0:           13.87        13.87        14.19         1.51
99.0:           32.16        32.16        32.48         0.66
99.7:           38.46        40.90        45.50         6.99
99.9:           48.70        49.34        51.90         3.34
worst:          67.97        67.71        74.88         6.59
----------------------------------------------------------------------------------------------------
-------------------------------- SUMMARY (UnsafeText.asDouble) us ----------------------------------
Percentile   run1         run2         run3      % Variation
50.0:            2.77         2.78         2.84         1.33
90.0:            2.90         2.91         2.97         1.45
99.0:            3.06         3.05         3.20         3.05
99.7:           17.18        16.37        17.57         4.66
99.9:           22.82        22.11        23.33         3.54
worst:          41.54        44.22        42.82         2.15
----------------------------------------------------------------------------------------------------
```


src/test/java/net/openhft/chronicle/core/MathsTest.java
```diff
@@ -462,4 +462,12 @@ public void floorNceilN() {
             assertEquals("i: " + i, floor0, floor, 0);
         }
     }
+
+    @Test
+    public void asDouble() {
+        assertEquals(12.345, Maths.asDouble(12345, 0, false, 3), 0.00000000000001);
+        assertEquals(-12.345, Maths.asDouble(12345, 0, true, 3), 0.00000000000001);
+        assertEquals(98760.0, Maths.asDouble(12345, 3, false, 0), 0.00000000000001);
+        assertEquals(1543.125, Maths.asDouble(12345, -3, false, 0), 0.0000000000001);
```
nicktindall, line 471: Some very rudimentary tests. The places it's used are also tested.

src/main/java/net/openhft/chronicle/core/util/StringUtils.java
```diff
@@ -377,7 +381,8 @@ private static boolean compareRest(@NotNull CharSequence in,
         return true;
     }
 
-    private static double asDouble(long value, int exp, boolean negative, int decimalPlaces) {
+    @Deprecated(/* For removal in x.25 */)
+    public static double asDouble(long value, int exp, boolean negative, int decimalPlaces) {
```
nicktindall: Just expose this one for the purpose of the comparison

src/main/java/net/openhft/chronicle/core/io/UnsafeText.java
```diff
@@ -249,7 +249,11 @@ private static long appendText(long address, String s) {
         return address;
     }
 
-    private static double asDouble(long value, int exp, boolean negative, int deci) {
+    /**
+     * @deprecated Use {@link Maths#asDouble(long, int, boolean, int)} instead
+     */
+    @Deprecated(/* To be moved into Maths.asDouble in x.25 */)
+    public static double asDouble(long value, int exp, boolean negative, int deci) {
```
nicktindall: Just expose so we can keep the history for now, delegate from Maths.asDouble (with additional zero-cost assertions)

src/main/java/net/openhft/chronicle/core/util/StringUtils.java
JerryShea: can we remove in x.24?

JerryShea: i.e. as part of this PR?

src/main/java/net/openhft/chronicle/core/Maths.java
```diff
@@ -762,4 +764,53 @@ public static int hash(Object o1, Object o2, Object o3, Object o4) {
...
+     * Convert components to a double value
+     *
+     * @param value         The integer value
+     * @param exponent      The exponent
+     * @param negative      Whether it is negative or not
+     * @param decimalPlaces The number of decimal places
+     * @return The value as a double
+     */
+    @SuppressWarnings("deprecation")
+    public static double asDouble(@NonNegative long value, int exponent, boolean negative, @NonNegative int decimalPlaces) {
```
JerryShea: this used to be in UnsafeText

JerryShea (approved):
LGTM but I would prefer if @peter-lawrey could also check

PR #487, opened 2023-02-27
Add ExceptionKey#containsText, Fixes #486

### LINKED ISSUE #486: Implement deep, null safe containsText method for ExceptionKey
We often match exceptions using message sub-strings, but the method for determining whether an ExceptionKey contains the specified string is repeated in every XXXTestCommon and often shallow and not null-safe. ExceptionKey should be able to tell you if it contains a sub-string.

nicktindall:


nicktindall:
Actually, this belongs in ExceptionTracker

PR #489, opened 2023-03-14
make Closeable#closeQuietly warn if Exception thrown

JerryShea:


src/main/java/net/openhft/chronicle/core/io/Closeable.java
```diff
@@ -46,7 +46,7 @@ static void closeQuietly(@Nullable Object o) {
             try {
                 ((java.lang.AutoCloseable) o).close();
             } catch (Exception e) {
-                Jvm.debug().on(Closeable.class, e);
+                Jvm.warn().on(Closeable.class, e);
             } catch (Throwable e) {
```
nicktindall, line 50: You could remove the catch exception block altogether because it's the same as the catch throwable one below it now?

nicktindall (approved):

src/main/java/net/openhft/chronicle/core/io/Closeable.java
JerryShea: yes for sure. Maybe we consider adding a message:
```
            try {
                ((java.lang.AutoCloseable) o).close();
            } catch (Throwable e) {
                Jvm.warn().on(Closeable.class, "Problem calling close(). Ignoring", e);
            }
```

nicktindall:
Can we go further (see https://github.com/OpenHFT/Chronicle-Core/pull/496)? the change in AbstractCloseable really needs to go hand in hand with this change

JerryShea:
Replaced by #496

PR #496, opened 2023-03-28
Don't swallow exceptions in close methods

nicktindall:
This goes further than removing the swallowing of exceptions in `Closeable#closeQuietly()`, but also stops swallowing them in `AbstractCloseable#close()`.


JerryShea (approved):
Agree 100%. We should never be catching Throwable and only logging at debug level. Once we have some more approvals, can we create an issue to document the change for release notes?

PR #500, opened 2023-03-30
Fix flaky JitterSamplerTest#takeSnapshot, deprecate JitterSampler#sleepSilently

nicktindall:
This one has failed in TC a few times, I assume because putting a `Thread.sleep` in the `takeSnapshot` loop saw it sleep through the whole "finishing" phase sometimes. I replaced it with a call to `Jvm.busyWaitMicros(1000)` which should never pause for longer than the expected time?

Also deprecated `JitterSampler#sleepQuietly` as it seemed to be a lesser `Jvm.pause(..)`

### DIFF
--- a/src/test/java/net/openhft/chronicle/core/threads/JitterSamplerTest.java
+++ b/src/test/java/net/openhft/chronicle/core/threads/JitterSamplerTest.java
@@ -21,45 +21,50 @@ package net.openhft.chronicle.core.threads;
 import net.openhft.chronicle.core.Jvm;
 import org.junit.Test;
 
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
 import static org.junit.Assert.*;
 
 public class JitterSamplerTest {
 
     @Test
     public void takeSnapshot() throws InterruptedException {
-
-//        Assume.assumeTrue(!Jvm.isArm());
+        CountDownLatch latch = new CountDownLatch(2);
         Thread t = new Thread(() -> {
             JitterSampler.atStage("started");
+            waitForLatch(latch);
             int millis = Jvm.isArm() ? 120 : 60;
-            JitterSampler.sleepSilently(millis);
+            Jvm.pause(millis);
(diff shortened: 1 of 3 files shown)


src/test/java/net/openhft/chronicle/core/threads/JitterSamplerTest.java
```diff
@@ -21,36 +21,30 @@
...
-        STARTED:
-        {
-            for (int j = 20; j >= 0; j--) {
-                Jvm.pause(20);
-                if (JitterSampler.desc != null)
-                    break STARTED;
-            }
-            fail("Not started");
-        }
+        waitForLatch(latch);
```
nicktindall, line 44: Use CountdownLatch so we don't need to busy wait

src/test/java/net/openhft/chronicle/core/threads/JitterSamplerTest.java
```diff
@@ -21,45 +21,50 @@
...
         for (int i = 0; i < 10; i++) {
-            JitterSampler.sleepSilently(10);
+            Jvm.busyWaitMicros(1000);
             String s = JitterSampler.takeSnapshot(10_000_000);
-            //System.out.println(s);
-            if ("finishing".equals(JitterSampler.desc)) {
+            final String desc = JitterSampler.desc;
```
nicktindall, line 48: Cache the value at the first moment it's read, I notice it also failed sometimes because it had gone from `started` to `finishing` between the `"finishing".equals(desc)` and the `assertEquals("started", desc)`

nicktindall: e.g. https://teamcity.chronicle.software/buildConfiguration/Chronicle_BuildAll_Build/791947?expandBuildProblemsSection=true&hideProblemsFromDependencies=false&expandBuildTestsSection=true&hideTestsFromDependencies=false

JerryShea (approved):

PR #501, opened 2023-03-31
Update EventHandler JavaDoc to more accurately represent the current API

nicktindall:
The JavaDoc (and contract) had a few inconsistencies and was probably a bit too verbose in places.

One issue was OpenHFT/Chronicle-Threads#160

It's not possible to always call `loopFinished` from the event loop thread if we guarantee it's always called (e.g. if the handler was never started there may be NO event loop thread to call `loopFinished` from).

I've changed the contract to indicate that if and only if `loopStarted` is called, `loopFinished` will be called, and `loopFinished` will always be called from the event loop thread.

The lifecycle of an event handler becomes:
- eventLoop(...) {once, on any thread}
    - loopStarted() {once, on eventloop thread}
    - action() {zero or more times, on eventloop thread}
    - loopFinished() {once, on eventloop thread}
- close() 

The indented calls either all happen or none happen

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/threads/EventHandler.java
+++ b/src/main/java/net/openhft/chronicle/core/threads/EventHandler.java
@@ -24,6 +24,7 @@ import java.io.Closeable;
 
 @FunctionalInterface
 public interface EventHandler extends VanillaEventHandler {
+
     /**
      * This method is called once when it is added to an eventLoop, which might be before the EventLoop has started
      * This could be called in any thread.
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/core/threads/EventHandler.java
```diff
@@ -34,28 +35,31 @@ default void eventLoop(EventLoop eventLoop) {
...
+     * <p>
+     * If this handler implements {@link Closeable} then the event loop will call close (once only) on this after
      * loopFinished has been called.
-     * <p>If this implements {@link Closeable} and something other than the event loop
-     * calls close then it is expected that this will throw {@link InvalidEventHandlerException} next time
-     * {@link #action()} is called. This will then allow the event loop to call loopFinished (and close) on this.
-     * If this use case is required it is strongly recommend that this event handler guards against close
-     * being called more than once (and ignores subsequent calls).
-     * <p>Exceptions thrown by loopFinished or close are caught and logged (at debug level)
-     * and cleanup continues.
```
nicktindall, line 58: This seems like an implementation detail and not a part of the contract we need to explain. If someone closes a handler, it's likely it'll throw some kind of exception when action is called, and we'll remove the handler. That much seems implicit in the contract.

src/main/java/net/openhft/chronicle/core/threads/EventHandler.java
```diff
@@ -34,28 +35,31 @@ default void eventLoop(EventLoop eventLoop) {
...
      * removed from the event loop.
-     * <p>If this implements {@link Closeable} then the event loop will call close (once only) on this after
+     * <p>
+     * If this handler implements {@link Closeable} then the event loop will call close (once only) on this after
      * loopFinished has been called.
-     * <p>If this implements {@link Closeable} and something other than the event loop
-     * calls close then it is expected that this will throw {@link InvalidEventHandlerException} next time
-     * {@link #action()} is called. This will then allow the event loop to call loopFinished (and close) on this.
-     * If this use case is required it is strongly recommend that this event handler guards against close
-     * being called more than once (and ignores subsequent calls).
```
nicktindall, line 56: This seems like an implementation detail and not a part of the contract we need to explain. If someone closes a handler, it's likely it'll throw some kind of exception when action is called, and we'll remove the handler. That much seems implicit in the contract.

JerryShea: this was added to make best practice explicit

peter-lawrey (approved):

Issue #497, opened 2023-03-29
GetSpecificationVersion can sometime return null and throw an NPE, which was not being caught


RobAustin:
When using vaNPE with the following version:

![image](https://user-images.githubusercontent.com/6506056/228583388-20fc388e-5d42-4260-a649-3f9e0fa9680a.png)

Stacktrace:
Exception in thread "main" java.lang.ExceptionInInitializerError
        at net.openhft.chronicle.core.UnsafeMemory.<clinit>(UnsafeMemory.java:73)
        at com.**.<…>
Caused by: java.lang.NullPointerException
        at net.openhft.chronicle.core.Bootstrap.getMajorVersion0(Bootstrap.java:112)
        at net.openhft.chronicle.core.Bootstrap.<clinit>(Bootstrap.java:69)
        ... 21 more

dpisklov:
It's actually not fixed unfortunately - the fix has an issue in it. Because `Bootstrap#getMajorVersion0()` is called from Jvm class initializer, accessing Jvm.warn() throws NPE as well - there either must be no logging or logging should use System.err.println

PR #511, opened 2023-05-17
Changed log at  System.err.println("Unable to get the major vers..") because JVM is not yet initialised 

RobAustin:
releates to issue #497


peter-lawrey (approved):

src/main/java/net/openhft/chronicle/core/Bootstrap.java
```diff
@@ -114,7 +114,7 @@ private static int getMajorVersion0() {
         try {
             return Integer.parseInt(Runtime.class.getPackage().getSpecificationVersion().split("\\.")[1]);
         } catch (Exception e) {
-            Jvm.warn().on(Jvm.class, "Unable to get the major version, defaulting to 8 " + e);
+            System.err.println("Unable to get the major version, defaulting to 8 " + e.getMessage());
```
peter-lawrey, line 117: Can you leave just + e)

PR #518, opened 2023-05-25
OS.mapAlign is not working correctly when offset is not already aligned.

gortiz:
I've found some segmentation faults when trying to integrate Chronicle-Core in Apache Pinot.

It seems that the problem is actually in mapAlign method, which rounds up to the next page (like `pageAlign` does). TBH I don't find specially useful the ability to map a random offset if I'm not going to know the position of my offset in the actually paged page. Instead I would prefer to receive an error if the offset is not page aligned or return a structure with the memory address and the offset inside that.

Anyway, I've added a test that verifies the behavior of `mapAlign`. As said, the javadoc is not very clear to me, so maybe the old behavior is what you actually expect. In that case we can revert the changes on OS.java and update OSTest.java accordenly to have a test that verifies the expected behavior.

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/OS.java
+++ b/src/main/java/net/openhft/chronicle/core/OS.java
@@ -285,7 +285,11 @@ public final class OS {
      */
     public static long mapAlign(long offset) {
         int chunkMultiple = (int) mapAlignment();
-        return (offset + chunkMultiple - 1) / chunkMultiple * chunkMultiple;
+        return mapAlign(offset, chunkMultiple);
+    }
+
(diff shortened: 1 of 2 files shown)


JerryShea:
Thanks @gortiz !

Would you be able to sign the [CLA](https://chronicle.software/contributor-agreement/) please?

gortiz:
I think I already signed that when doing a previous contribution in Chronicle-Bytes, but I already did it again just in case it was actually needed.

JerryShea:
Fantastic thanks @gortiz !

PR #520, opened 2023-05-29
Method to check whether running in core event loop

### LINKED ISSUE #519: Add method to check whether running in core event loop
Some of event loop users may need such check to make sure their methods are only called in thread safe fashion

alamar:
closes #519

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/threads/EventLoop.java
+++ b/src/main/java/net/openhft/chronicle/core/threads/EventLoop.java
@@ -84,4 +84,13 @@ public interface EventLoop extends Closeable {
      */
     @Override
     void close();
+
+    /**
+     * Returns {@code true} if current thread is the core thread of this event loop, or if it cannot be determined.
+     * If current thread is not the core event thread but any other, returns {@code false}.
+     * @return {@code true} unless current thread is not the core event thread
+     */
+    default boolean runsInsideCoreLoop() {
+        return true;
+    }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/core/threads/EventLoop.java
```diff
@@ -84,4 +84,13 @@ static boolean inEventLoop() {
      */
     @Override
     void close();
+
+    /**
```
JerryShea, line 88: maybe this should be in `net.openhft.chronicle.threads.CoreEventLoop`?

alamar: We want this to have as high in interface hierarchy as possible to minimize amount of casting. EventLoop exists in 99% cases to be a proxy for CoreEventLoop anyway.

JerryShea (approved):

src/main/java/net/openhft/chronicle/core/threads/EventLoop.java
```diff
@@ -84,4 +84,13 @@ static boolean inEventLoop() {
...
+     * Returns {@code true} if current thread is the core thread of this event loop, or if it cannot be determined.
+     * If current thread is not the core event thread but any other, returns {@code false}.
+     * @return {@code true} unless current thread is not the core event thread
+     */
+    default boolean runsInsideCoreLoop() {
+        return true;
```
JerryShea, line 94: I think default value should be false

alamar: Imagine somebody has an alternate event loop and we broke their code by introducing this check - I think it should be true because of that

PR #524, opened 2023-06-05
reduce verbosity of license check message if >30 days to run

JerryShea:


src/main/java/net/openhft/chronicle/core/LicenceCheck.java
```diff
@@ -34,32 +32,46 @@ public interface LicenceCheck {
...
+        licenceExpiry(product, caller, logLicenceExpiryDetails);
+    }
+
+    /**
+     * Provide licence expiry details
+     * @param product product
+     * @param caller caller
+     * @param licenceExpiryDetails callback to call with license days to run and license owner
+     */
+    static void licenceExpiry(String product, Class<?> caller, BiConsumer<Long, String> licenceExpiryDetails) {
```
JerryShea, line 64: for https://github.com/ChronicleEnterprise/Chronicle-FIX/issues/1687

alamar: Does it add property checking? I don't see that

alamar (approved):

src/main/java/net/openhft/chronicle/core/LicenceCheck.java
JerryShea: It doesn't. This change a) reduces verbosity and b) provides the licenseExpiry method so a consumer can find out how many days remaining

alamar, line 64: Sonar says caller arg may be removed

JerryShea: Yes but it is also an argument to the `check` method and so passed through. Assume we may use called in the future

PR #529, opened 2023-06-12
Avoid incomplete cleanup in case of no specific thread

alamar:


JerryShea (approved):

peter-lawrey (commented):
Looks like an improvement

PR #534, opened 2023-06-21
withstand extra characters e.g. newline in license file. Remove old comment

JerryShea:


src/main/java/net/openhft/chronicle/core/LicenceCheck.java
```diff
@@ -67,7 +67,7 @@ static void licenceExpiry(String product, Class<?> caller, BiConsumer<Long, Stri
             String expiryDateFile = product + ".expiry-date";
             try {
                 String source = new String(IOTools.readFile(LicenceCheck.class, expiryDateFile));
-                LocalDate expiryDate = LocalDate.parse(source);
+                LocalDate expiryDate = LocalDate.parse(source.substring(0, 10));
```
nicktindall: Should we just do `source.trim()` that would presumably cover off the case where there is whitespace before the date too?

nicktindall (approved):

src/main/java/net/openhft/chronicle/core/LicenceCheck.java
JerryShea: yes good idea, and cleaner code too

PR #535, opened 2023-06-22
don't use finaliser by default

JerryShea:


peter-lawrey (approved):
We should add a note somewhere that if there is a bug that only shows up in production, this will result in a resource leak.

nicktindall (approved):
This one concerns me a bit because I fear if we accidentally leave it off, tests will no longer break when we're not managing resources properly. We need to be sure to turn this on in the tests in every repo.

JerryShea:
Good point, I will do that

Issue #549, opened 2023-07-14
Timed out waiting for the Finalizer on Zing


alamar:
Probably started springing up after #531

Seems to only happen on Zing VMs.

PR #550, opened 2023-07-14
Multiple GC attempts to make Zing happy #549

alamar:


alamar:
Tested with Chronicle-FIX to ensure the Finalizer problem no longer appears

JerryShea (approved):
Approved but maybe add a comment explaining why we do this

PR #555, opened 2023-08-03
Update OS.java

JohannesLichtenberger:
The switch doesn't work on Windows. I don't know if removing it is the proper solution, though... may have been added for a reason ;-)


GylesBates:
This code looks to be just removing the java 20 block which was presumably added for some reason. Instead we do an if-statement based on Windows or unix and for windows use FileDispatcherImpl instead which seems to be the windows equivilent implementation of FileDispatcher. I will submit a pull request for this

PR #564, opened 2023-08-17
JMH for TimeProviders

JerryShea:


benchmarks/pom.xml
```diff
@@ -132,7 +132,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
-                <version>2.2</version>
+                <version>3.2.1</version>
```
nicktindall: Are we able to remove the version (is it set in third-party?)

nicktindall (approved):

rogersimmons (approved):

benchmarks/pom.xml
JerryShea: done

PR #572, opened 2023-08-30
Handle compact strings disabled

peter-lawrey:


src/main/java/net/openhft/chronicle/core/util/StringUtils.java
```diff
@@ -352,6 +360,8 @@ private static void ensureJava9Plus() {
     @Java9
     @NotNull
     public static String newStringFromBytes(byte @NotNull [] bytes) {
+        if (!HAS_ONE_BYTE_PER_CHAR)
```
JerryShea, line 363: I appreciate that branch prediction is incredibly effective, but would it not be more JIT-friendly if instead of this conditional we created a static final `Function<String, byte[]>' in the static block and called that?

RobAustin (approved):

src/main/java/net/openhft/chronicle/core/util/StringUtils.java
peter-lawrey: While this function can be inlined, it still has to check the class of the function to ensure it is a method that has been inlined. However, `static final boolean` can be inlined even non final boolean can be inlined and the code eliminated for the branch not taken.

PR #574, opened 2023-08-31
Explicitly ban unsafe package for serialization

peter-lawrey:


src/main/java/net/openhft/chronicle/core/pool/ClassAliasPool.java
```diff
@@ -190,6 +192,22 @@ private Class<?> doLookup(String name) {
         }
     }
 
+    private boolean banned(String name) {
+        if (name.isEmpty())
```
JerryShea, line 196: why would name be empty?

peter-lawrey: It shouldn't, that's why it's banned

JerryShea: should we throw an exception or assert instead?

JerryShea (approved):

src/main/java/net/openhft/chronicle/core/pool/ClassAliasPool.java
peter-lawrey: I could leave this code out, but you would get an IndexOutOfBoundException which could be confusing.

peter-lawrey: All banned class names result in a specific ClassNotFoundRuntimeException

PR #576, opened 2023-09-19
Scoped thread locals

### LINKED ISSUE #575: Add "scoped" thread-locals
There is a pattern that is repeated in our code whereby we use singleton thread-local resources to avoid creating garbage.

Some examples include
- [Wires.acquireBinaryWire](https://github.com/OpenHFT/Chronicle-Wire/blob/d0e7b1c47f84a62a91eefe9f4f1a49d7b49e0acd/src/main/java/net/openhft/chronicle/wire/Wires.java#L394)
- [Wires.acquireBytes](https://github.com/OpenHFT/Chronicle-Wire/blob/d0e7b1c47f84a62a91eefe9f4f1a49d7b49e0acd/src/main/java/net/openhft/chronicle/wire/Wires.java#L372)
- [Wires.acquireAnotherBytes](https://github.com/OpenHFT/Chronicle-Wire/blob/d0e7b1c47f84a62a91eefe9f4f1a49d7b49e0acd/src/main/java/net/openhft/chronicle/wire/Wires.java#L403)
- [Wires.acquireBytesForToString](https://github.com/OpenHFT/Chronicle-Wire/blob/d0e7b1c47f84a62a91eefe9f4f1a49d7b49e0acd/src/main/java/net/openhft/chronicle/wire/Wires.java#L382)
- [WireInternal.acquireStringBuilder](https://github.com/OpenHFT/Chronicle-Wire/blob/d0e7b1c47f84a62a91eefe9f4f1a49d7b49e0acd/src/main/java/net/openhft/chronicle/wire/WireInternal.java#L108)
- [WireInternal.acquireAnotherStringBuilder](https://github.com/OpenHFT/Chronicle-Wire/blob/d0e7b1c47f84a62a91eefe9f4f1a49d7b49e0acd/src/main/java/net/openhft/chronicle/wire/WireInternal.java#L113)
- [WireInternal.acquireStringBuilderForValueIn](https://github.com/OpenHFT/Chronicle-Wire/blob/d0e7b1c47f84a62a91eefe9f4f1a49d7b49e0acd/src/main/java/net/openhft/chronicle/wire/WireInternal.java#L326)
- [WireInternal.acquireStringBuilderForValueOut](https://github.com/OpenHFT/Chronicle-Wire/blob/d0e7b1c47f84a62a91eefe9f4f1a49d7b49e0acd/src/main/java/net/openhft/chronicle/wire/WireInternal.java#L330)
- [WireInternal.acquireInternalBytes](https://github.com/OpenHFT/Chronicle-Wire/blob/d0e7b1c47f84a62a91eefe9f4f1a49d7b49e0acd/src/main/java/net/openhft/chronicle/wire/WireInternal.java#L317)

As you can see they have proliferated because sometimes to avoid a usage-collision, being thread-local is not enough. Indeed there have been a few bugs that have been the result of a thread using it's thread-local bytes in an outer scope, then the same thread over-writing it in an inner scope.

One way to mitigate this risk is for threads to take a lease on these resources and then return them once they're done with them. While the lease is being held the resource should not be available to any other nested scope.

The proposed pattern is:

```java

try (ScopedResource<Bytes<?>> stlBytes = Wires.acquireScopedBytes()) {
    // Use the bytes for whatever (we are holding an exclusive lease on it)
    Bytes<?> bytes = stlBytes.get();
    bytes.append("Hello world: ");
    bytes.append(callSubMethodThatMightAlsoTryAndUseTheThreadLocal());
    valueOut.text(bytes);
} // <-- the bytes instance is returned to the "pool" and made available to other callers again
```

Following this approach would eliminate a class of errors, and worst-case, when the pool is not large enough, generate some garbage, though this should also generate a warning so the usage pattern can be reviewed.

nicktindall:
Fixes #575

Also use a scoped thread-local for the `StringBuilderPool`

Kept the interfaces generic enough that we could investigate more elaborate pools, e.g ones that were thread-safe and shared instances between threads.

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/io/Wget.java
+++ b/src/main/java/net/openhft/chronicle/core/io/Wget.java
@@ -19,6 +19,8 @@
 package net.openhft.chronicle.core.io;
 
 import net.openhft.chronicle.core.pool.StringBuilderPool;
+import net.openhft.chronicle.core.scoped.ScopedResource;
+import net.openhft.chronicle.core.scoped.ScopedResourcePool;
 
 import java.io.BufferedReader;
 import java.io.IOException;
@@ -33,7 +35,7 @@ import static net.openhft.chronicle.core.io.Closeable.closeQuietly;
  */
 public final class Wget {
 
-    private static final StringBuilderPool STRING_BUILDER_POOL = new StringBuilderPool();
+    private static final ScopedResourcePool<StringBuilder> STRING_BUILDER_POOL = StringBuilderPool.createThreadLocal();
 
     private Wget() {
     }
@@ -44,11 +46,24 @@ public final class Wget {
      * @param url the URL of the HTTP GET request
      * @return the result of the request as a {@link CharSequence}
      * @throws IOException if an error occurs while establishing the connection
+     * @deprecated Use {@link #url(String, StringBuilder)} instead, this implementation is very dangerous
      */
+    @Deprecated(/* To be removed in x.26 */)
     public static CharSequence url(String url) throws IOException {
+        try (final ScopedResource<StringBuilder> sbTl = STRING_BUILDER_POOL.get()) {
+            url(url, sbTl.get());
+            return sbTl.get();
+        }
+    }
 
-        final StringBuilder sb = STRING_BUILDER_POOL.acquireStringBuilder();
-
+    /**
+     * Performs an HTTP GET request to the specified URL.
+     *
+     * @param url the URL of the HTTP GET request
+     * @param sb  The StringBuilder to which the response will be written to
+     * @throws IOException if an error occurs while establishing the connection
+     */
+    public static void url(String url, StringBuilder sb) throws IOException {
         InputStream is = null;
         try {
             is = new URL(url).openStream();
@@ -59,10 +74,8 @@ public final class Wget {
                     sb.append(s);
                 }
             }
-            return sb;
         } finally {
             closeQuietly(is);
         }
     }
-
 }
--- a/src/main/java/net/openhft/chronicle/core/pool/StringBuilderPool.java
+++ b/src/main/java/net/openhft/chronicle/core/pool/StringBuilderPool.java
@@ -18,6 +18,10 @@
 
 package net.openhft.chronicle.core.pool;
 
+import net.openhft.chronicle.core.Jvm;
+import net.openhft.chronicle.core.scoped.ScopedResourcePool;
+import net.openhft.chronicle.core.scoped.ScopedThreadLocal;
+
 import static java.lang.ThreadLocal.withInitial;
 
 /**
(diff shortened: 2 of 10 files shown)


src/main/java/net/openhft/chronicle/core/io/Wget.java
```diff
@@ -44,11 +44,22 @@ private Wget() {
...
      * @return the result of the request as a {@link CharSequence}
      * @throws IOException if an error occurs while establishing the connection
+     * @deprecated Use {@link #url(String, StringBuilder)} instead
      */
     public static CharSequence url(String url) throws IOException {
-
         final StringBuilder sb = STRING_BUILDER_POOL.acquireStringBuilder();
+        url(url, sb);
+        return sb;
+    }
```
nicktindall, line 57: This is a bit nasty, returning the thread-local singleton string builder instance

src/main/java/net/openhft/chronicle/core/pool/StringBuilderPool.java
```diff
@@ -39,10 +59,23 @@ public class StringBuilderPool {
...
+    /**
+     * Returns a scoped, thread-local StringBuilder instance from the pool.
+     * If the current thread does not yet have a StringBuilder, it is created and added to the pool.
+     * The length of the StringBuilder is reset to 0 before being returned.
+     *
+     * @return a StringBuilder instance with length 0.
+     */
+    public ScopedResource<StringBuilder> acquireStringBuilderScoped() {
+        return scopedStringBuilderPool.get();
+    }
```
nicktindall, line 83: We could potentially deprecate the whole StringBuilderPool and just use a ScopedResourcePool instead

src/main/java/net/openhft/chronicle/core/scoped/ScopedResourcePool.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public interface ScopedResourcePool<T> {
+
+    /**
+     * Get a scoped instance of the shared resource
+     *
+     * @return the {@link ScopedResource}, to be closed once it is finished being used
+     */
+    ScopedResource<T> get();
+}
```
nicktindall, line 26: Create this generic interface so we can implement e.g. a thread-safe one that shares instances across threads.

src/main/java/net/openhft/chronicle/core/pool/StringBuilderPool.java
JerryShea: works for me

JerryShea (approved):

src/main/java/net/openhft/chronicle/core/pool/StringBuilderPool.java
nicktindall: Done, `StringBuilderPool` will just provide a factory method going forward `StringBuilderPool#createThreadLocal`

PR #578, opened 2023-10-06
Add support for hugetlbfs (custom pageSize).

yevgenp:
Required for Chronicle Queue hugetlbfs support

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/OS.java
+++ b/src/main/java/net/openhft/chronicle/core/OS.java
@@ -42,6 +42,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
 import static java.lang.management.ManagementFactory.getRuntimeMXBean;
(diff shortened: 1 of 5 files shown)


tgd:
Hi Yevgen - I *think* it looks good - is the idea to introduce this new `mapAlignForPageSize` across the different call sites in bytes / wire etc? Does the `pageSize` method need any changes?

yevgenp:
> Hi Yevgen - I _think_ it looks good - is the idea to introduce this new `mapAlignForPageSize` across the different call sites in bytes / wire etc? Does the `pageSize` method need any changes?

Yes, that's the idea. I've already done it for Bytes. I need to release this first.

tgd:
Hi @yevgenp thanks for putting this together. I think its heading in the right direction.

yevgenp:
@peter-lawrey I've adjusted current PR with #579. I've removed new method `mapAlignForPageSize` (though I believe the naming was better), cleaned up OS validations a bit. If you don't mind I'd prefer to keep this one.
Please rereview/approve.

PR #580, opened 2023-10-10
Use FileDispatcherImpl over base class UnixFileDispatcherImpl 

GylesBates:
Currently you cannot use OS class in java 21 in windows due to UnixFileDispatcherImpl not existing.

I've looked at the various JDK implementations for windows/linux/unix/mac-os in the jdk and they all have a FileDispatcherImpl but the unix varients (mac-os/linux/unix) have an intermediately class "UnixFileDispatcherImpl" which is currently referenced. 

Apologies if pull request is into wrong branch - wasn't sure if it should go into develop or ea

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/OS.java
+++ b/src/main/java/net/openhft/chronicle/core/OS.java
@@ -59,7 +59,7 @@ public final class OS {
         try {
             Method map0;
             if (Jvm.isJava20Plus()) {
-                Class<?> dispatcherClass = findClass("sun.nio.ch.UnixFileDispatcherImpl");
+                Class<?> dispatcherClass = findClass("sun.nio.ch.FileDispatcherImpl");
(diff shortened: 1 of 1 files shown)


JerryShea:
@GylesBates I just took the liberty of changing to develop as that is where changes are supposed to land. We don't currently have a Java 21 Windows build in our CI to test this with. We expect to have full Java 21 support finished in the next week or two so once that is done then maybe you could rebase your branch and re-test? @peter-lawrey FYI

GylesBates:
Hi @JerryShea, thanks for the reply. Apologies was unaware of correct branch. I'm going to close this and create a new pull request from a fork of the develop branch instead.

JerryShea:
@GylesBates it is possible just to change the destination branch in the PR and then (if necessary) rebase your branch on that (although this should not be necessary as develop and ea are generally the same - part of Chronicle's release process in our TeamCity builds all the different repositories develop branches and when they all pass together, they are merged from develop to ea so we can cut releases from ea)

PR #582, opened 2023-10-17
Changes to compile in windows on java 21

GylesBates:
Continuation of pull request #580 

There were a few changes needed to get this compiling - notably the java.lang.Compiler seems to not be present in java 21 anymore. it was flagged as deprecated before so I guess they have officially removed it.

Second, after speaking with some colleages it appears that not every OS build of the jdk will have FileDispatcherImpl so we've put an if-windows type check.

This is compiling and tests passing locally for me.

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/Jvm.java
+++ b/src/main/java/net/openhft/chronicle/core/Jvm.java
@@ -477,11 +477,7 @@ public final class Jvm {
      */
     public static void nanoPause() {
         if (onSpinWaitMH == null) {
-            if (isJava9Plus())
-                Safepoint.force(); // 1 ns on Java 11
-            else
-                //noinspection removal
-                Compiler.enable(); // 5 ns on Java 8
+            Safepoint.force();
         } else {
             try {
                 onSpinWaitMH.invokeExact();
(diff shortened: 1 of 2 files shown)


JerryShea:
Thanks @GylesBates !

I don't believe that Chronicle-Core is designed to compile on Java > some number greater than 8, but @peter-lawrey can confirm

GylesBates:
> Thanks @GylesBates !
> 
> I don't believe that Chronicle-Core is designed to compile on Java > some number greater than 8, but @peter-lawrey can confirm

If i understand you correctly, what you're saying is you would still want to compile chronicle-core against java 11 or some earlier version and hence we could put back in those calls to Compiler but leave in the fix to resolve the map0 method resolution? I'm happy to take those changes out of the pull request if that is the case, it was mainly because i was reproducing issue by running the unit tests.

src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -477,11 +477,7 @@ public static void pause(final long durationMs) {
      */
     public static void nanoPause() {
         if (onSpinWaitMH == null) {
-            if (isJava9Plus())
-                Safepoint.force(); // 1 ns on Java 11
-            else
-                //noinspection removal
-                Compiler.enable(); // 5 ns on Java 8
+            Safepoint.force();
```
peter-lawrey, line 480: Unfortunately, that will mean it doesn't work on Java 8.

PR #584, opened 2023-10-18
Change Jvm.nanoPause() to compile on Java 21

peter-lawrey:


yevgenp (approved):

src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -922,15 +923,12 @@ public static void addSignalHandler(final SignalHandler signalHandler) {
 
     /**
      * Inserts a low-cost Java safe-point in the code path.
+     * <p>
+     * Note: this does nothing on Java 8.
```
JerryShea, line 927: @peter-lawrey this is a breaking change. I don't think this is a good idea

src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -922,15 +923,12 @@ public static void addSignalHandler(final SignalHandler signalHandler) {
 
     /**
      * Inserts a low-cost Java safe-point in the code path.
```
JerryShea, line 925: "only if SAFEPOINT_ENABLED"

PR #587, opened 2023-10-19
Add a safepoint for Java 8 that compiles in Java 21

peter-lawrey:


peter-lawrey:
Compile.enable() was faster but doesn't compile in Java 21.
Snapshot.force() didn't work on older versions of Java 8, but from version 1.8.0_202 it's fine (possibly older)
Snapshot.force() takes around 16 ns on a Ryzen 7 5950X

src/main/java/net/openhft/chronicle/core/Jvm.java
```diff
@@ -922,9 +920,7 @@ public static void addSignalHandler(final SignalHandler signalHandler) {
     }
 
     /**
-     * Inserts a low-cost Java safe-point in the code path.
-     * <p>
-     * Note: this does nothing on Java 8.
+     * Inserts a low-cost Java safe-point in the code path if -Djvm.safepoint.enabled
```
JerryShea, line 924: we should also edit comment in https://github.com/OpenHFT/Chronicle-Core/blob/885750bdab180c35a6e61b8822b85eb7a537a9e5/src/main/java/net/openhft/chronicle/core/Jvm.java#L1536

JerryShea (approved):
Approved subject to one additional change (to a comment) required. Good to hear that Safepoint.force works on Java 8 too

src/main/java/net/openhft/chronicle/core/Jvm.java
peter-lawrey: This has been checked to work on Java 8 update 202+, removed the comment

PR #592, opened 2023-10-30
Fix deprecated classes for x.25

peter-lawrey:


src/main/java/net/openhft/chronicle/core/Mocker.java
```diff
@@ -0,0 +1,157 @@
...
+import java.io.StringWriter;
+import java.util.concurrent.BlockingQueue;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+
+/**
+ * The Mocker class provides utility methods for creating mocked instances of interfaces.
+ */
+// TODO to be moved to the util package x.25
+@Deprecated(/* to be removed in x.25 */)
```
yevgenp, line 34: Did you mean 2.26?

yevgenp (approved):

src/main/java/net/openhft/chronicle/core/Mocker.java
peter-lawrey: We will see.

PR #611, opened 2024-01-04
Deprecate StringUtils.setCount as not compatible with Java 21.

yevgenp:


src/main/java/net/openhft/chronicle/core/util/StringUtils.java
```diff
@@ -330,6 +330,7 @@ public static byte[] extractBytes(@NotNull String s) {
         return getMemory().getObject(s, S_VALUE_OFFSET);
     }
 
+    @Deprecated(/* remove in x.26 */)
     public static void setCount(@NotNull StringBuilder sb, int count) {
         getMemory().writeInt(sb, SB_COUNT_OFFSET, count);
```
JerryShea, line 335: should instead call `StringBuilder.setLength(count)`?

yevgenp: It's not that straightforward. Bytes relies on this implementation. So it's gonna stay for now.
I don't intend to merge it yet.

JerryShea: yep - I see it is an optimisation to avoid some setLength side-effects

yevgenp:
Will not be deprecated as it's still being used in `Bytes` (to be moved ?)
It's **UNSAFE** to use it alone.

PR #612, opened 2024-01-08
Disable DirectBufferUtilTest.cleanIfInstanceOfDirectBufferShouldCleanDirectBuffer in Java9+.

yevgenp:
Breaks the BuildAll. See email thread.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -230,6 +230,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.12.1</version>
                 <configuration>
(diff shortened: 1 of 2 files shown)


tgd (approved):

pom.xml
```diff
@@ -230,6 +230,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.12.1</version>
```
nicktindall, line 233: Why did we need to put versions of the plugins in here? I believe we strive to manage those in the `root-parent-pom`

JerryShea: Yep - we should not be adding these

PR #603, opened 2023-11-24
Remove deprecated method, fix issues

alamar:
closes #602

PR #615, opened 2024-01-10
Prepare copyMemory for removal in x.26

peter-lawrey:


tgd (approved):

src/main/java/net/openhft/chronicle/core/Memory.java
```diff
@@ -378,7 +378,7 @@ default long heapUsed() {
          * @param length  the number of bytes to copy
          * @deprecated This method is deprecated and will be removed in a future version (x.24).
          */
-    @Deprecated(/* to be removed in x.24 */)
+    @Deprecated(/* to be removed in x.26 */)
```
alamar, line 381: I'm not sure this will help. You cannot remove this method without removing the rest of copyMemory() ones, since a different overload will be quietly used which has different behavior (offset for object header v.s. not offset).
See https://github.com/OpenHFT/Chronicle-Core/pull/603

src/main/java/net/openhft/chronicle/core/UnsafeMemory.java
```diff
@@ -1156,7 +1160,7 @@ public void copyMemory(@Nullable Object src, long srcOffset, @Nullable Object de
             if (dest instanceof byte[]) {
                 copyMemory((byte[]) src, Math.toIntExact(srcOffset - Unsafe.ARRAY_BYTE_BASE_OFFSET), (byte[]) dest, Math.toIntExact(destOffset - ARRAY_BYTE_BASE_OFFSET), length);
             } else {
-                copyMemoryLoop(src, ARRAY_BYTE_BASE_OFFSET + Math.toIntExact(srcOffset - Unsafe.ARRAY_BYTE_BASE_OFFSET), dest, destOffset, length);
+                copyMemoryLoop(src, ARRAY_BYTE_BASE_OFFSET + Math.toIntExact(srcOffset), dest, destOffset, length);
```
alamar, line 1163: How would existing code work in light of this change?

PR #632, opened 2024-02-29
Support converting to a class alias

peter-lawrey:


src/main/java/net/openhft/chronicle/core/pool/ClassLookup.java
```diff
@@ -97,4 +97,32 @@ default ClassLookup wrap(@NotNull ClassLoader classLoader) {
      * @param names A single alias or a comma-separated string of aliases for the class.
      */
     void addAlias(Class<?> clazz, String names);
+
+    /**
+     * Applies an alias transformation to the given class name if an alias exists. This method
```
tgd, line 102: Great Javadoc :+1: 

Is this something that should get an entry in the README @peter-lawrey ?

tgd (approved):

PR #636, opened 2024-03-18
improve doco for SingleThreadedCheck

JerryShea:


README.adoc
```diff
@@ -205,7 +208,7 @@ boolean isWindows = OS.isWindows();
 boolean is64bit = OS.is64Bit();
 String hostname = OS.getHostName();
 String username = OS.getUserName();
-String targetDir = OS.getTarget(); // where is the target directory during builds.
+String targetDir = OS.getTarget(); // location the target directory during builds
```
nicktindall, line 211: location **of** the target directory?

README.adoc
```diff
@@ -475,6 +475,62 @@ To make sure the shutdown hook does not prevent classes from unloading, deregist
 | X | `false` | Resources are not queued and are released synchronously.
 |===
 
+== Thread Safety Checks
+
+Classes that are designed for single-threaded use can implement `net.openhft.chronicle.core.io.SingleThreadedChecked`.
+There are a number of implementations of this in Chronicle-Core, including `net.openhft.chronicle.core.io.AbstractCloseable`,
+and many Chronicle library classes extend this class. When the user calls a method which _must_ be called single-threaded, these methods call a private method to check that `Thread.currentThread()` is the same as the first thread that used the object, and throws an exception if not. `AbstractCloseable` also provides very helpful functionality (if resource tracing is on) to keep track of the stack trace of where it was used first by the use-by thread - this provides invaluable when diagnosing unexpected sharing of an object between threads.
```
nicktindall, line 482: This is quite a long line, should we put a newline in for people reading it un-rendered without wrap turned on? I think Ctrl-Alt-L does something like that.

nicktindall: Also

> this provides invaluable

should be proves?

README.adoc
```diff
@@ -475,6 +475,62 @@ To make sure the shutdown hook does not prevent classes from unloading, deregist
...
+
+Thread safety checking can be turned off with the system property `disable.single.threaded.check`, once the application is thoroughly tested.
+
+An object can be handed-off between threads by calling `singleThreadedCheckReset`.
```
nicktindall, line 486: Is it worth going over the different use-cases where you might trip over them and how to avoid? In my experience those are:
- when one thread is finished with a resource and handing it to another - `singleThreadedCheckReset`
- when multiple threads access a resource, but there is external mechanism to prevent concurrent access `singleThreadedCheckDisable`
- when you want us to get out of your damn way because you know what you're doing `disable.single.threaded.check`

nicktindall (approved):

PR #637, opened 2024-03-22
Error on deprecated cleanup

peter-lawrey:


tgd (approved):

src/main/java/net/openhft/chronicle/core/util/ByteBuffers.java
```diff
@@ -27,10 +27,7 @@
  * Utility class for working with {@link ByteBuffer} instances.
  * <p>
  * Provides functionality to directly set the address and capacity of a ByteBuffer.
- * <p>
- * Note: This class is deprecated and will be removed in version x.26.
  */
-@Deprecated(/* to be removed x.26 */)
```
JerryShea: why has this been undeprecated?

src/main/java/net/openhft/chronicle/core/util/Ints.java
```diff
@@ -32,7 +32,6 @@
  * A collection of functional compositions to check and assert various requirements
  * related to {@code int} values.
  */
-@Deprecated(/* to be moved in x.26 */)
```
JerryShea, line 35: why move deprecated tags to functions?

src/main/java/net/openhft/chronicle/core/util/Longs.java
```diff
@@ -68,6 +61,28 @@ public static long requireNonNegative(final long value) {
...
+     * public Foo(long bar) {
+     *     this.bar = requirePositive(bar);
+     * }
+     * </pre></blockquote>
+     *
+     * @param value the value to check
+     * @return the provided {@code value} if the check passes
+     * @throws IllegalArgumentException If the check fails
+     */
+    public static long requirePositive(final long value) {
```
JerryShea, line 80: some of these are being kept?

src/main/java/net/openhft/chronicle/core/util/Ints.java
peter-lawrey: A few functions are used many times, but most aren't used at all.

src/main/java/net/openhft/chronicle/core/util/Longs.java
peter-lawrey: the ones that were used, yes.

src/main/java/net/openhft/chronicle/core/util/ByteBuffers.java
peter-lawrey: The class is needed in one place. So I have moved to that place on review.

PR #638, opened 2024-03-27
X.26

yevgenp:


yevgenp:
@peter-lawrey please check the deprecations left. There some dependencies thus I was unable to remove them.

src/main/java/net/openhft/chronicle/core/util/Longs.java
```diff
@@ -33,7 +33,6 @@
  * A collection of functional compositions to check and assert various requirements
  * related to {@code long} values.
  */
-@Deprecated(/* to be moved to internal in Chronicle-Bytes in x.26 as that is where it's used. One method is used in UnsafeMemory */)
```
JerryShea: if we are going to de-deprecate classes/methods then this change should also be made in another PR and merged ASAP into develop. That way, people won't remove calls to methods which are no longer deprecated

peter-lawrey: done

tgd (approved):

peter-lawrey (approved):

src/main/java/net/openhft/chronicle/core/Mocker.java
```diff
@@ -21,11 +21,8 @@
...
 
 import java.io.PrintStream;
-import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.concurrent.BlockingQueue;
-import java.util.function.BiConsumer;
-import java.util.function.Consumer;
 
 /**
  * The Mocker class provides utility methods for creating mocked instances of interfaces.
```
JerryShea, line 28: should refer to util.Mocker with a comment explaining this here left for backwards compatibility?

JerryShea (approved):
approved with 1 comment

src/main/java/net/openhft/chronicle/core/Mocker.java
peter-lawrey: I have added a comment

PR #640, opened 2024-03-28
Instead of removing the whole class, remove the method that aren't us…

peter-lawrey:
…ed by other modules.


JerryShea (approved):
This is great - thanks @peter-lawrey

PR #641, opened 2024-04-01
review with lint:all

peter-lawrey:


src/main/java/net/openhft/chronicle/core/internal/Bootstrap.java
```diff
@@ -186,4 +186,9 @@ private static int getProcessId0() {
         System.err.println(Bootstrap.class.getName() + ": Unable to determine PID, picked 1 as a PID");
         return rpid;
     }
+
+    @SuppressWarnings("unchecked")
+    public static <T> T uncheckedCast(Object obj) {
```
yevgenp, line 191: I'd name it just `cast`. It's concise and still clear enough.

yevgenp (approved):

src/main/java/net/openhft/chronicle/core/internal/Bootstrap.java
peter-lawrey: Having it longer makes it less likely people will use it.
It's not intended to be a nice solution but a workaround. 
(Otherwise, ordinarily I would agree that something concise would be nicer)

PR #645, opened 2024-04-08
Cache resolution failures to avoid repeated attempts

### LINKED ISSUE #644: In DumpMain, class resolving takes 2/3 of execution time when queue value type not present
Probably elsewhere. It turns out that Class.forName() is an expensive operation and should not be done repeatedly. Let us try caching the absense of the value to avoid redoing this lookup on every entry.

alamar:
closes #644

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/pool/ClassAliasPool.java
+++ b/src/main/java/net/openhft/chronicle/core/pool/ClassAliasPool.java
@@ -38,13 +38,12 @@ import java.util.concurrent.TimeUnit;
  */
 public class ClassAliasPool implements ClassLookup {
     public static final ClassAliasPool CLASS_ALIASES = new ClassAliasPool(null).defaultAliases();
-    private static final ClassNotFoundRuntimeException cachedEx = new ClassNotFoundRuntimeException(
-            new ClassNotFoundException("Cached resolution failure, will not retry"));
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/core/pool/ClassAliasPool.java
```diff
@@ -146,22 +148,39 @@ public Class<?> forName(@NotNull CharSequence name) throws ClassNotFoundRuntimeE
...
         return forName0(key);
     }
 
     @NotNull
     private synchronized Class<?> forName0(@NotNull CAPKey key) throws ClassNotFoundRuntimeException {
         Class<?> clazz = stringClassMap2.get(key);
-        if (clazz != null) return clazz;
+        if (clazz != null) {
+            if (clazz == ResolutionFailure.class)
+                throw cachedEx;
```
JerryShea: shouldn't we throw a new exception with the class details filled in?

nicktindall: Yeah I think having a class name in there would make troubleshooting a lot easier if it needed to be done. Perhaps it would mean a lot of object creation with all the exceptions. I suppose you could cache those too ;)

alamar: I have a better idea that I will implement, thank you for the food for thoughs.

JerryShea (approved):
LGTM!

PR #669, opened 2024-09-19
Check empty methods and asciidoc fix

peter-lawrey:


src/test/java/net/openhft/chronicle/core/JvmTest.java
```diff
@@ -271,6 +272,7 @@ static class ClassD extends ClassC {
     }
 
     interface InterfaceWithDefaultMethod {
+        @SuppressWarnings("EmptyMethod")
```
JerryShea, line 275: should this be on helloDefault?

src/test/java/net/openhft/chronicle/core/JvmTest.java
```diff
@@ -460,9 +462,11 @@ static class DTO {
     @AnnoAlias
     interface Foo {
 
+        @SuppressWarnings("EmptyMethod")
```
JerryShea: required on an interface?

tgd (changes requested):
Just need to check the suppress warnings empty method on interfaces to see if required

src/test/java/net/openhft/chronicle/core/JvmTest.java
peter-lawrey: The warning is that all implementations are empty.

tgd (approved):

PR #684, opened 2025-01-04
StringUtils.equalsCaseIgnore optimisation fixes #683

### LINKED ISSUE #683: StringUtils.equalsCaseIgnore optimisation
`StringUtils.equalsCaseIgnore` can be optimised for the common case where both strings exactly match.

tgd:
StringUtils.equalsCaseIgnore can be optimised for the common case where both strings exactly match. See #683 

`equalsCaseIgnore` is used in wire during deserialisation - `net.openhft.chronicle.wire.TextWire#read(java.lang.CharSequence, int, java.lang.Object)`

Screenshot shows before and after:

image

### DIFF
--- a/benchmarks/src/main/java/net/openhft/chronicle/core/benchmark/StringUtilsEqualsCaseIgnoreJLBH.java
+++ b/benchmarks/src/main/java/net/openhft/chronicle/core/benchmark/StringUtilsEqualsCaseIgnoreJLBH.java
@@ -1,0 +1,62 @@
+package net.openhft.chronicle.core.benchmark;
+
+import net.openhft.chronicle.core.Jvm;
+import net.openhft.chronicle.core.util.StringUtils;
+import net.openhft.chronicle.jlbh.JLBH;
+import net.openhft.chronicle.jlbh.JLBHOptions;
+import net.openhft.chronicle.jlbh.JLBHTask;
+import net.openhft.chronicle.jlbh.TeamCityHelper;
+
+public class StringUtilsEqualsCaseIgnoreJLBH implements JLBHTask {
+
+    private static String input;
+    private final int iterations;
+    private JLBH jlbh;
+
+    private StringUtilsEqualsCaseIgnoreJLBH(int iterations) {
+        this.iterations = iterations;
+    }
+
+    @Override
+    public void init(JLBH jlbh) {
+        this.jlbh = jlbh;
+    }
+
+    @Override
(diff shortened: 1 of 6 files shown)


benchmarks/src/main/java/net/openhft/chronicle/core/benchmark/StringUtilsEqualsCaseIgnoreJLBH.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public void run(long startTimeNS) {
+        StringUtils.equalsCaseIgnore(input, input);
+        jlbh.sample(System.nanoTime() - startTimeNS);
+    }
+
+    @Override
+    public void complete() {
+        TeamCityHelper.teamCityStatsLastRun(this.getClass().getSimpleName(), jlbh, iterations, System.out);
```
JerryShea: maybe consider a similar approach to `net.openhft.chronicle.bytes.perf.ContentEqualJLBH` where a few different lengths and and best/worst cases tried

src/main/java/net/openhft/chronicle/core/util/StringUtils.java
```diff
@@ -253,19 +253,23 @@ private static boolean isEqualJava9(@NotNull StringBuilder s, @NotNull CharSeque
...
-        for (int i = 0; i < cs.length(); i++)
-            if (Character.toLowerCase(charAt(s, i)) !=
-                    Character.toLowerCase(charAt(cs, i)))
+    public static boolean equalsCaseIgnore(@Nullable CharSequence s1, @NotNull CharSequence s2) {
+        if (s1 == null) return false;
+        if (s1.length() != s2.length()) return false;
+        for (int i = 0; i < s2.length(); i++) {
+            char c1 = charAt(s1, i);
+            char c2 = charAt(s2, i);
+            if (c1 == c2)
```
JerryShea, line 268: nice

JerryShea (approved):

benchmarks/src/main/java/net/openhft/chronicle/core/benchmark/StringUtilsEqualsCaseIgnoreJLBH.java
tgd: Good idea @JerryShea I have set up benchmarks for best and worst cases; both are looking good

image

PR #685, opened 2025-01-11
Improve StackTrace after review of generated user guide

peter-lawrey:


tgd (approved):

src/main/java/net/openhft/chronicle/core/StackTrace.java
```diff
@@ -69,46 +116,60 @@ public StackTrace(String message, Throwable cause) {
...
     /**
-     * A lightweight version of {@code StackTrace} that avoids filling in the stack trace.
+     * Converts the current (nano-precision) time from the system CLOCK to a {@link ZonedDateTime}.
+     *
+     * @return the current time as a {@link ZonedDateTime}, with nanosecond precision
+     */
+    @NotNull
+    private static ZonedDateTime nanosAsZonedDateTime() {
+        long nowNanos = CLOCK.currentTimeNanos();
+        int nanosPerSecond = 1_000_000_000;
```
benbonavia: minor: this can be a constant

benbonavia (approved):

src/main/java/net/openhft/chronicle/core/StackTrace.java
peter-lawrey: Good suggestion

PR #688, opened 2025-01-29
Source time for stack trace timestamp from system time

chroniclekevinpowe:


src/main/java/net/openhft/chronicle/core/StackTrace.java
```diff
@@ -134,7 +133,8 @@ public static StackTrace forThread(Thread t) {
      */
     @NotNull
     private static ZonedDateTime nanosAsZonedDateTime() {
-        long nowNanos = CLOCK.currentTimeNanos();
+        // temporary change to confirm effect - only creates time at millisecond precision
+        long nowNanos = System.currentTimeMillis() * NANOS_PER_MILLI;
```
benbonavia, line 137: could we not use `System.nanoTime()`?

src/test/java/net/openhft/chronicle/core/StackTraceTest.java
```diff
@@ -40,38 +36,24 @@ static void thinking() {
...
     public void testDefaultConstructor() {
         StackTrace st = new StackTrace();
         String currentThreadName = Thread.currentThread().getName();
-        assertEquals(
-                "stack trace on " + currentThreadName
-                        + " at 2024-01-02T03:04:05.006007008Z",
-                st.getMessage()
+        assertTrue(
+                String.format("%s must match regular expression expecting 'stack trace on %s at' with following timestamp", st.getMessage(), currentThreadName),
+                st.getMessage().matches("stack trace on " + currentThreadName + " at \\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d+Z$")
```
benbonavia: this regex pattern can be a constant

src/main/java/net/openhft/chronicle/core/StackTrace.java
chroniclekevinpowe: Unfortunately not.

src/test/java/net/openhft/chronicle/core/StackTraceTest.java
chroniclekevinpowe: Good point - ta

benbonavia (approved):

PR #689, opened 2025-01-29
Make timestamp inclusion in stack trace optional

chroniclekevinpowe:
Default stack trace generation to not include stack trace, but allow optional inclusion of timestamp in stack trace.

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/StackTrace.java
+++ b/src/main/java/net/openhft/chronicle/core/StackTrace.java
@@ -80,7 +80,17 @@ public class StackTrace extends Throwable {
      * default message "stack trace".
      */
     public StackTrace() {
-        this("stack trace");
+        this("stack trace", false);
+    }
+
+    /**
+     * Constructs a new {@code StackTrace} for the current thread with the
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/core/StackTrace.java
```diff
@@ -80,7 +80,15 @@ public class StackTrace extends Throwable {
...
     public StackTrace() {
-        this("stack trace");
+        this("stack trace", false);
+    }
+
+    /**
+     * Constructs a new {@code StackTrace} for the current thread with the
+     * default message "stack trace", optionally including a timestamp in the stack trace message.
+     */
+    public StackTrace(boolean addTimestamp) {
```
benbonavia, line 92: maybe use the `@param` tag to describe the purpose of `addTimestamp` for the constructors for more strict javadoc correctness

chroniclekevinpowe: Ta - I missed one, clearly! More speed, less haste. :-)

benbonavia (approved):

PR #760, opened 2025-05-27
documentation-only sweep across net.openhft.chronicle.core.onoes.

peter-lawrey:
This pull request is a **documentation-only** sweep across *net.openhft.chronicle.core.onoes*.  
It finishes the work started in #751-#759, then removes a few items that review feedback flagged as noise.

| Patch | Theme | Notes |
|-------|-------|-------|
| 01/13 | **Whitespace** | Strip stray blanks from `package-info.java` to stop AsciiDoc paragraph drift. |
| 02-10/13 | **Javadoc overhaul** | Add or rewrite comments for every public type:<br>• crisp first-sentence summaries<br>• full `@param/@return/@throws` sets<br>• explicit thread-safety statements<br>• SLF4J mapping table in `LogLevel`<br>• examples and `@implNote/@apiNote`s<br>• modernised package overview. |
| 11-12/13 | **Noise removal** | Delete boiler-plate `{@inheritDoc}` blocks introduced during earlier iterations – the methods are self-explanatory. |
| 13/13 | **Tag hygiene** | Drop the `@since 3.25ea` markers that were making the docs verbose |

### Why we need it
* **Cleaner generated docs** – IDE tool-tips and the public site no longer italicise parameter names or show duplicated text.  
* **Accurate contracts** – Users now see thread-safety guarantees and fallback behaviour explicitly called out.  
* **Maintenance** – The codebase no longer carries superfluous tags, so future contributors have less clutter to maintain.  

There are **no code-path or binary changes**; all unit tests still pass.

### Impact & migration
Zero runtime impact – consumers do **not** need to recompile or change any code. The only visible change is richer, cleaner documentation.

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/onoes/package-info.java
+++ b/src/main/java/net/openhft/chronicle/core/onoes/package-info.java
@@ -27,7 +27,6 @@
  * }</pre>
  *
  * @author OpenHFT - part of Chronicle-Core
- * @since 3.25ea
  * @see net.openhft.chronicle.core.onoes.ExceptionHandler
  */
 package net.openhft.chronicle.core.onoes;
--- a/src/main/java/net/openhft/chronicle/core/onoes/package-info.java
+++ b/src/main/java/net/openhft/chronicle/core/onoes/package-info.java
@@ -5,12 +5,10 @@
  * in a uniform manner throughout an application. This includes mechanisms for
  * chaining multiple exception handlers together, logging exceptions, handling
  * exceptions on a per-thread basis, and recording exceptions.
- * 
  * <p>
  * The core of this package is the {@link net.openhft.chronicle.core.onoes.ExceptionHandler}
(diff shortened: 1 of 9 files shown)


src/main/java/net/openhft/chronicle/core/onoes/package-info.java
```diff
@@ -1,32 +1,32 @@
...
- * 
  *
+ * Example usage for switching to {@link net.openhft.chronicle.core.onoes.RecordingExceptionHandler}:
+ * <pre>{@code
+ * Map<ExceptionKey, Integer> counts = Jvm.recordExceptions();
+ * // run code that generates exceptions
+ * Jvm.resetExceptionHandlers();
+ * }</pre>
+ *
+ * @author OpenHFT - part of Chronicle-Core
```
tgd: author tag

tgd (changes requested):
Spotted an author tag in the last javadoc

PR #812, opened 2025-06-01
Harden `Wget`, add `LimitedInputStream`, and modernise the test-suite

peter-lawrey:
### Pull Request — **Harden `Wget`, add `LimitedInputStream`, and modernise the test-suite**

This series of eight commits turns the one-liner `Wget.url(…)` helper into a small, well-tested, security-oriented component:

| ✔                | Area                                                                                                                                                                                                 | Summary |
| ---------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
| **API**          | *Builder pattern, injectable seams* — `ConnectionProvider` & `CharsetDetector` can be stubbed, so callers and tests stay offline and deterministic.                                                  |         |
| **Security**     | *Hard byte cap* via the new **`LimitedInputStream`** (package-private). EOF vs. overflow is now unambiguous.                                                                                         |         |
| **Resilience**   | Full guard-rail set:<br>• URL-length check (2 048)<br>• Scheme allow-list (http/https)<br>• Negative/zero budget handling<br>• Time-outs applied to the default connection                           |         |
| **Performance**  | Streamed `BufferedReader` read with 4 KiB buffer (commit 2/8).                                                                                                                                       |         |
| **Docs & Style** | Javadoc everywhere, uniform headers, explanatory commit messages.                                                                                                                                    |         |
| **Tests**        | 40+ assertions across *smoke* and *edge-case* suites, covering concurrency, misbehaving collaborators, EOF semantics, and reflection-verified time-outs. All green on Java 8+ (`mvn test` ≈ 120 ms). |         |

---

#### 🚀  What’s new?

* **`LimitedInputStream`**

  * Drop-in, zero-dependency wrapper that enforces a byte budget.<br>
    • returns `-1` when both wrapper *and* source hit EOF.
    • throws `IOException` (“Size limit exceeded”) if the source still has data.

* **`Wget` refactor**

  * **Builder API** with sane defaults, optional overrides.
  * Default provider applies `connectTimeout` & `readTimeout` (tests prove this).
  * Charset auto-detection hook; null-safe UTF-8 fallback.
  * Buffered 4 KiB char copying (faster than single-char loop).
  * Explicit URL-length guard and scheme allow-list.
  * Negative `maxResponseBytes` rejected early.

* **Formatting / documentation**

  * Consistent 80-column Javadoc paragraphs, ordered imports, UTF-8 header dashes, explicit section rulers.

---

#### 🔍  Test coverage highlights

* **Happy-path**: body read, custom charset, builder façade.
* **Edge-cases**:

  * zero, exact, and overflowing budgets
  * provider & appendable exceptions
  * misbehaving `CharsetDetector`
  * unknown `Content-Length`, never-ending stream
  * reflection-verified connect/read time-outs
  * 20-thread re-entrancy stress
* **Regression fix**: correct appendable-exception path (commit 3/8).

---

#### ⚠️  Breaking or behavioural changes

| Old                                                                                             | New                                                                                                   |
| ----------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------- |
| `Wget.url(String, StringBuilder)` silently followed redirects, no length guard, read all bytes. | Now enforces byte cap (default 10 MiB), **no redirects**, URL length ≤ 2 048, and throws on overflow. |
| Single static method, hard to stub.                                                             | Builder + injectable seams; default façade still present for back-compat.                             |
| No time-outs.                                                                                   | 10 s **connect/read** defaults, configurable per builder.                                             |

Applications that relied on unlimited downloads or redirects need to opt-in by supplying a custom `ConnectionProvider`.

---

#### 🧩  How to migrate

```java
// Old one-liner
StringBuilder sb = new StringBuilder();
Wget.url("https://example.com", sb);

// New preferred style
StringBuilder sb = new StringBuilder();
new Wget.Builder()
        .connectTimeoutMs(5_000)
        .readTimeoutMs(5_000)
        .maxResponseBytes(2L << 20)   // 2 MiB
        .build()
        .fetch("https://example.com", sb);
```

The original static façade still works, but now benefits from the security defaults.


src/main/java/net/openhft/chronicle/core/io/LimitedInputStream.java
```diff
@@ -0,0 +1,83 @@
...
+ *   <li>If the wrapped stream is at true end‑of‑file, {@code -1} is returned (normal EOF).</li>
+ *   <li>If additional data is still available, an {@link IOException} with the message
+ *       <q>Size limit exceeded</q> is thrown.  This ensures the caller can never read past
+ *       the configured limit.</li>
+ * </ul>
+ *
+ * <p>The class is package‑private on purpose; use it through public APIs such as
+ * {@link Wget} instead of referencing it directly.</p>
+ */
+final class LimitedInputStream extends FilterInputStream {
```
tgd, line 23: Recommend adding a unit test for this

src/main/java/net/openhft/chronicle/core/io/Wget.java
```diff
@@ -1,59 +1,118 @@
...
+ * <p>{@code Wget} is intentionally simple and dependency‑free.  It opens a connection to
+ * an {@code http://} or {@code https://} URL, honours user‑defined time‑outs, and streams
+ * at most <em>N</em> bytes into a supplied {@link Appendable}.  Everything that touches
+ * the outside world can be replaced via the {@link Builder} for predictable tests.</p>
  */
 public final class Wget {
 
-    private Wget() {
-    }
+    /* ───────── Functional seams ───────── */
```
tgd: What does seams mean in this context?

tgd (approved):

src/main/java/net/openhft/chronicle/core/io/Wget.java
peter-lawrey: Fixed

src/main/java/net/openhft/chronicle/core/io/LimitedInputStream.java
peter-lawrey: Added

PR #823, opened 2025-07-22
Enhance GenericReflection to handle TypeVariable bounds and add test for parameter types

peter-lawrey:
This pull request improves the way `GenericReflection.getParameterTypes(...)` handles generic parameters declared with bounds (e.g. `A extends Number`). Instead of returning the raw `TypeVariable` instance, it now resolves the parameter to its first bound (or `Object.class` if no explicit bounds are present). Additionally, the recursive collection of superclasses and interfaces in `getGenericClassesSuperclassesAndInterfaces(...)` is made cleaner by casting the input `Type` to `Class<?>` once up front (removing repeated casts). These changes ensure more accurate runtime type resolution and cleaner code. 

## Background

Reflective APIs such as `Method.getGenericParameterTypes()` often return `TypeVariable` objects for parameters declared with generics, which can obscure the actual concrete types at runtime.
The `TypeVariable.getBounds()` method returns the array of upper bounds declared for a type variable, defaulting to `Object.class` if none are specified.
Many frameworks normalize generic bounds by using the first bound or falling back to `Object` to simplify downstream processing and avoid dealing with raw type variables.

## Changes

1. **Resolve bounded type variables to their first bound**

   * In `getParameterTypes(...)`, after calling `findType(...)`, check if the resolved `Type` is a `TypeVariable<?>`. If so, extract `typeVariable.getBounds()[0]` or `Object.class` when no bounds exist 
   * This replaces raw `TypeVariable` occurrences with concrete classes like `Number` or `CharSequence` for `A extends Number` or `B extends CharSequence`, respectively.

2. **Consolidate class casting in recursive type collection**

   * In `getGenericClassesSuperclassesAndInterfaces(Type forClass, ...)`, cast `forClass` to `Class<?>` once (`Class<?> forClass2`) instead of repeating `(Class)` each time 
   * This reduces redundant casts and clarifies that only non-generic `Class` types are traversed here.

3. **Add dedicated unit test for bounded generics**

   * Introduce `getParameterTypesExtends()` in `GenericReflectionTest` to verify that parameters declared as `<A extends Number, B extends CharSequence>` are resolved to `[Number.class, CharSequence.class]` 

## Impact

* **Improved type accuracy** when reflecting on methods that use generics with bounds—users no longer need to manually unwrap `TypeVariable` instances to their bounds.
* **Better compatibility** with frameworks and libraries that expect concrete `Class<?>` types rather than `TypeVariable` objects during reflective operations.
* **Minimal behavioral change** for methods without generic bounds, since the first bound defaults to `Object.class`, matching previous expectations when no bounds were declared

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/util/GenericReflection.java
+++ b/src/main/java/net/openhft/chronicle/core/util/GenericReflection.java
@@ -99,7 +99,18 @@ public enum GenericReflection {
     public static Type[] getParameterTypes(Method method, Type type) {
         final Type[] parameterTypes = method.getGenericParameterTypes();
         for (int i = 0; i < parameterTypes.length; i++) {
-            parameterTypes[i] = findType(method, type, parameterTypes[i]);
+            Type typeI = findType(method, type, parameterTypes[i]);
+            if (typeI instanceof TypeVariable) {
+                TypeVariable<?> typeVariable = (TypeVariable<?>) typeI;
+                Type[] bounds = typeVariable.getBounds();
+                if (bounds.length > 0) {
+                    typeI = bounds[0];
+                } else {
+                    // If no bounds, use Object as a fallback
+                    typeI = Object.class;
+                }
+            }
+            parameterTypes[i] = typeI;
         }
         return parameterTypes;
     }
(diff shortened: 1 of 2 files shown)


peter-lawrey:
Running a Build-All bumped on this

james-mcsherry (approved):

PR #838, opened 2025-12-18
SLF4J logging now is not linked to debug being enabled

benbonavia:
all `isEnabled(class)` calls were linked to whether the debug logger was enabled, revert to correct this issue


chroniclekevinpowe (approved):
Approved. Thank you for catching this!

chroniclekevinpowe:
Could we potentially add a test in Core to prevent a similar regression?

PR #842, opened 2026-03-16
Cache thread liveness once per orphan-cleanup sweep

peter-lawrey:
## Summary

Reduce repeated `Thread.isAlive()` calls during `CleaningThreadLocal` orphan cleanup by caching liveness once per distinct thread for the duration of one complete cleanup sweep.

Let:

* `E` be the total number of tracked thread/value entries across all `CleaningThreadLocal` instances.
* `U` be the number of distinct tracked threads.

The sweep still visits all `E` entries, but the number of liveness checks is reduced from `E` to `U` when thread identities are shared across locals.

## Implementation

* Create one short-lived `IdentityHashMap<Thread, Boolean>` after the empty-set check.
* Share that cache while draining every tracked `CleaningThreadLocal` in the sweep.
* Discard the cache after the sweep so liveness is checked again on the next invocation.
* Preserve the existing separation between draining under the global lock and running cleanup callbacks outside it.
* Keep the public API unchanged.

Thread identities are unique within each individual `nonCleaningThreadValues` map. Cache reuse occurs across different `CleaningThreadLocal` instances populated by the same threads.

## Behavioural impact

Cleanup remains best-effort. Liveness is sampled once per distinct thread during a sweep. If a thread is observed alive and then terminates part-way through that sweep, values encountered later for that thread may remain until the next sweep. A fresh cache is created for every sweep.

No long-lived state, expiry policy or additional synchronisation is introduced.

## Validation

The cross-local integration test uses the same four worker threads to populate five locals and verifies:

* four liveness checks per populated sweep rather than twenty;
* live values remain tracked;
* a later sweep refreshes liveness after the workers terminate;
* all twenty values are cleaned exactly once; and
* a subsequent empty sweep performs no repeated cleanup.

Relevant tests pass on Java 8 and Java 21, and the full Maven verification passes on Java 21.

## Performance evidence

The call-count test establishes that the corrected cache eliminates repeated liveness checks when threads are shared across locals. It does not establish an elapsed-time improvement. Any performance claim should be supported separately with before/after measurements on the JVM and workload associated with the reported timeout, including a no-sharing control where each thread appears in only one local.

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/threads/CleaningThreadLocal.java
+++ b/src/main/java/net/openhft/chronicle/core/threads/CleaningThreadLocal.java
@@ -11,6 +11,7 @@ import org.jetbrains.annotations.Nullable;
 
 import java.util.*;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.function.UnaryOperator;
 
@@ -216,8 +217,16 @@ public class CleaningThreadLocal<T> extends ThreadLocal<T> {
      *
      * <p>Call at whatever cadence suits your application (e.g.&nbsp;every few
      * seconds, once a minute, or only at JVM shutdown).</p>
+     *
+     * <p>Liveness is sampled once per distinct thread during each sweep. If a
+     * thread terminates after being observed alive, its remaining values are
+     * retained until the next sweep.</p>
      */
     public static void cleanupNonCleaningThreads() {
+        cleanupNonCleaningThreads(Thread::isAlive);
+    }
+
+    static void cleanupNonCleaningThreads(Predicate<Thread> isThreadAlive) {
         // Drain stale entries under the lock, then run user cleanup
         // callbacks outside it so slow cleanup code never blocks
         // set()/remove()/initialValue() across the JVM.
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/core/threads/CleaningThreadLocal.java
```diff
@@ -228,12 +228,16 @@ private boolean doCleanupNonCleaningThreads() {
         if (!trackNonCleaningThreads)
             return true;
 
+        // cache isAlive results so it doesn't become O(N * M) on a costly operation.
+        Map<Thread, Boolean> isThreadAlive = new IdentityHashMap<>();
```
tgd, line 242: Does this cache need to be hoisted up a level?

tgd (changes requested):

src/main/java/net/openhft/chronicle/core/threads/CleaningThreadLocal.java
peter-lawrey: The cache is short lived as it caches the current state of threads being alive, this is only valid for a short window

tgd (approved):

PR #843, opened 2026-03-16
Simplify waitForCloseablesToClose() to focus on waiting semantics

peter-lawrey:
Refactor `CloseableUtils.waitForCloseablesToClose(long)` so its behaviour matches its name and return type more clearly: it now only waits for tracked closeables to enter the closing state within the given timeout and returns `true` or `false` accordingly.

## Motivation

`waitForCloseablesToClose(long)` previously mixed two responsibilities:

* waiting for closeables to close within a timeout
* performing assertion-style validation and exception throwing on unreleased resources

That made the method harder to read and blurred the distinction between it and `assertCloseablesClosed()`, which is the method intended for strict validation and failure reporting.

This change makes the contract clearer:

* `waitForCloseablesToClose(long)` is a polling/wait helper
* `assertCloseablesClosed()` remains the diagnostic/assertion path

## Change

The method has been simplified to:

* invoke pending cleanup work up front

  * `CleaningThreadLocal.cleanupNonCleaningThreads()`
  * `BackgroundResourceReleaser.releasePendingResources()`
* check whether all tracked `ManagedCloseable` instances are already closing
* return `true` once they are
* return `false` if the timeout expires
* pause briefly between checks rather than re-running assertion-oriented logic

The call in `finalize()` has also been updated to use the local method directly, improving readability.

## What was removed

The previous implementation included behaviour that was more appropriate for assertion/debug flows, including:

* disabling single-threaded checks on `AbstractCloseable`
* calling `throwExceptionIfNotReleased()` on `ReferenceCountedTracer`
* retrying around `IllegalStateException`

That logic obscured the purpose of the method and overlapped with `assertCloseablesClosed()`.

## Behavioural intent

This is primarily a clarity and responsibility-separation change:

* `waitForCloseablesToClose(long)` now has straightforward wait semantics
* `assertCloseablesClosed()` remains the place for stronger verification and failure signalling

## Benefits

* clearer API intent
* simpler control flow
* easier distinction between “wait and report success/failure” and “assert and diagnose”
* reduced surprise from a boolean-returning method throwing assertion-related failures during polling

## Notes

A `@see` link has been added in both directions to make the relationship between the two methods explicit.


src/main/java/net/openhft/chronicle/core/internal/CloseableUtils.java
```diff
@@ -124,56 +124,47 @@ protected void finalize() throws Throwable {
      *
      * @param millis The time limit in milliseconds to wait for the closeable resources to close.
      * @return true if all closeable resources are closed within the time limit, false otherwise.
+     * @see #assertCloseablesClosed()
      */
     @SuppressWarnings({"java:S3776", "java:S3516"}) // turned on by assert
     public static boolean waitForCloseablesToClose(long millis) {
```
tgd, line 130: To what extent will this change semantics and impact upstream code? Have we got green bumped builds upstream for everything?

tgd (changes requested):

src/main/java/net/openhft/chronicle/core/internal/CloseableUtils.java
peter-lawrey: This is used in the shutdown of tests when checking their resources. This behaviour should be cleaner and clearer but potentially a bit slower as a result, ie it favours waiting a bit than pulling everything down. This could show up tests that currently sometimes fail on cleanup to more consistently fail.
A Build-All bumped passed

tgd (approved):

PR #849, opened 2026-04-23
Tighten `Jvm.parseSize` exact-byte semantics and pin fuzz-surfaced `StringUtils` behaviour

peter-lawrey:
This PR tightens `Jvm.parseSize(...)` so size strings are parsed with exact byte-count semantics, changes `Jvm.getSize(...)` so blank property values are treated as invalid input rather than silently unset, and documents/pins several fuzz-surfaced `StringUtils` edge cases with regression tests.

### `Jvm.parseSize(String)`

* Switched the numeric path from `Double.parseDouble(...)` + `Math.round(...)` to `BigDecimal`.
* Fractional sizes are now accepted **only** when they resolve to an exact whole-byte count.

  * Example: `0.5kb -> 512` still works.
  * Example: `0.1MB -> 104857.6` now throws instead of rounding to `104858`.
* Empty / blank input is rejected explicitly.
* Negative sizes are rejected explicitly.
* Uppercase `I` is now accepted alongside lowercase `i` in binary suffixes.
* Lone unit-prefix inputs such as `iB` / `IB` now fail with a clean `IllegalArgumentException` instead of `StringIndexOutOfBoundsException`.
* Trimming is applied after suffix stripping, so inputs like `1 ki B` are handled consistently with `1 kiB`.
* Added `@NonNegative` to reflect the method post-condition.
* Updated Javadoc:

  * corrected the `G/g` factor from `2^10` to `2^30`
  * replaced the old `0.001TiB -> 1.024 GiB` example with `0.25TiB -> 256 GiB`, which remains valid under exact-byte semantics

### `Jvm.getSize(String, long)`

* Blank property values are now forwarded to `parseSize(...)`, logged via `Jvm.warn(...)`, and then fall back to `defaultValue`.
* Only an absent (`null`) property continues to return the default silently.
* This makes blank misconfiguration visible instead of treating it the same as “unset”.

### `StringUtils`

* `StringUtils.parseDouble("")` now explicitly returns `0`.

  * This lenient behaviour is intentional and relied on by Chronicle-Bytes YAML / JSON parsing paths.
* `StringUtils.toTitleCase(...)` now uses effective case conversion rather than `Character.isLowerCase(...)` alone when detecting boundaries.

  * This makes case-stable Unicode letters behave more consistently and preserves idempotence.
* Clarified in Javadoc that `StringUtils.startsWith(...)` is case-insensitive, matching existing behaviour and `endsWith(...)`.

## Behaviour changes worth flagging

1. Size values that previously rounded silently now fail unless they map to an exact byte count.

   * For example, `0.1gb` will now fail instead of rounding.
   * Exact values such as `0.5kb`, `0.125MB`, and `0.25TiB` still work.

2. Blank system-property values that previously fell back silently in `Jvm.getSize(...)` now emit a warning before falling back.

   * Behaviour for absent (`null`) properties is unchanged.

## Tests

Added / updated coverage for:

* exact size parsing
* invalid size inputs
* blank property fallback behaviour
* empty-string `parseDouble(...)`
* `toTitleCase(...)` idempotence with case-stable letters
* case-insensitive `startsWith(...)` / `endsWith(...)`

New / expanded test coverage includes:

* `JvmParseSizeInvalidTest`
* `JvmParseSizeTest`
* `StringUtilsTest`

## Notes

* `Jvm.parseSize(...)` is reached via configuration parsing, not on a per-message hot path, so the `BigDecimal` allocations are not expected to affect steady-state runtime paths.
* The diff also includes minor formatting-only changes in `Jvm.java` outside the functional changes. These are IDE auto-format artefacts and can be reverted if preferred.

## For reviewers

Please focus review on:

* whether the move from rounding to exact-byte semantics in `Jvm.parseSize(...)` is the right compatibility tradeoff
* whether blank property values in `Jvm.getSize(...)` should now warn rather than be treated as silently unset
* whether the documented lenient `StringUtils.parseDouble("") -> 0` behaviour matches expected downstream usage
* whether the `toTitleCase(...)` Unicode/idempotence fix is the right boundary rule
* whether any downstream code may depend on the old fractional rounding behaviour and needs follow-up

If you want, I can also turn this into a shorter, more “Chronicle-style” PR description with less detail.


tgd (changes requested):
Hi @peter-lawrey - thank you - please could you link this to a jira - there are also some merge conflicts on the branch

peter-lawrey:
Covered in Jira https://chroniclesoftware.atlassian.net/browse/CORE-58

tgd (approved):

Issue #634, opened 2024-03-05
Catch Throwable instead of Exception


JohannesLichtenberger:
`net.openhft.chronicle.core.Jvm.maxDirectMemory0` should catch `Throwable` instead of `Exception`, otherwise I can't create a usable native image of my DBS project: https://github.com/oracle/graal/issues/8406#issuecomment-1979042559

tgd:
Hi @JohannesLichtenberger thanks for bringing this to our attention - I've had a quick look through the linked thread but can you please summarize why this change needs to be made / how it blocks use of native image?

We do read issues created here but I can't share a resolution timeline yet, the more info we have the better regarding how this may impact users. We offer commercial support which we prioritize - if you need to expedite that please let me know directly or here - https://chronicle.software/contact-us/

JohannesLichtenberger:
@tgd The issue is described here: https://github.com/oracle/graal/issues/8406#issuecomment-1963941154

JohannesLichtenberger:
@tgd it should be really easy to fix, plus I don't get any money with SirixDB (only spending money so far...) :-)

At some point however I'd be probably interested in spending some money on performance improvements, just because it's interesting to learn for me :)

tgd:
Hi @JohannesLichtenberger - I've taken a look and yes indeed changing this call site is not much work. However, there are plenty of other callsites where we catch Exception and not Throwable. Will you run into the same issue there?

In general we are not focused on supporting graal at this time - as mentioned in the other thread we are depending on a number of JVM internals which may well cause further issues running on SubstrateVM.

But let's use this issue to detail some of the requirements and we can figure out where this sits in our roadmap.

JohannesLichtenberger:
Sounds great. In any case, I'm also thinking about switching to `MemorySegments` or some lightweight abstraction on top, but I guess for now it would be easier to still use Chronicle Bytes (mostly I'm simply using `Bytes.allocateElasticOnHeap()`, so probably I won't run into other issues).

Would be wort a try IMHO if you could release a new patch version.

JerryShea:
hi @JohannesLichtenberger - I've made the change in [this branch](https://github.com/OpenHFT/Chronicle-Core/tree/issue/634-throwable). As @tgd says, there are other places that the same change may need to be made. Can I suggest you build that branch locally and try it out?

PR #850, opened 2026-04-24
Tighten over-broad exception catches across Chronicle Core and bundle minor utility cleanups

peter-lawrey:
This PR reviews over-broad exception handling across Chronicle Core and replaces generic `catch (Exception)` / `catch (Throwable)` clauses with the concrete failure types expected at each call site, wherever that can be done safely.

A small number of broad boundaries are intentionally retained where the code is acting as a last-resort cleanup or best-effort platform fallback. Those sites are now commented, so the rationale is explicit during future security or Sonar-driven cleanup passes.

Alongside that review, this PR also bundles a handful of adjacent, low-risk cleanups in the same files. Those are itemised separately below so they can be reviewed independently from the main exception-boundary work.

## Primary change: exception broadness review

* Narrow broad reflective, networking, filesystem, and optional-platform catches across:

  * `Jvm`
  * `OS`
  * `AnnotationFinder`
  * `Bootstrap`
  * `ReferenceCountedUtils`
  * `InternalPomProperties`
  * `ClassLookup`
  * `AbstractInvocationHandler`
  * `ObjectUtils`
* Keep broad handling only where it is intentional:

  * `AbstractReferenceCounted.inThreadPerformRelease()` now explicitly uses `catch (Throwable)` as the last-resort containment boundary for background release work.
  * `CleaningThreadLocal.cleanup(...)` now explicitly uses `catch (Throwable)` as the final cleanup boundary when tracked resource cleanup must continue even after serious failures.
  * `ObjectUtils.newInstanceOrNull(...)` keeps its broad null-on-failure contract, but now emits a debug log before returning `null`.
  * `MaxMemoryHolder.maxDirectMemory0()` broadens the catch after user feedback that this was causing a problem. #634 
* Tighten `Jvm.addToClassPath(...)` so it only swallows URL/URI-to-path conversion failures instead of unrelated throwables.
* Keep the optional JVM-internals fallback paths in `Jvm` best-effort: if direct-memory / reserved-memory introspection is unavailable, the existing degraded fallback remains.
* Make Java-version fallback parsing in `Bootstrap` more explicit by handling null or malformed specification-version strings before defaulting to Java 8.

## Bundled minor non-exception-review changes

* `Mocker`

  * return `ObjectUtils.defaultValue(...)` for ignored proxy methods so primitive-returning methods no longer fail by returning `null`
* `BackgroundResourceReleaser`

  * replace direct `Thread.sleep(1)` polling with `Jvm.pause(1)`
  * stop the wait loop if the current thread becomes interrupted while waiting for resource cleanup
* `PriorityHook`

  * snapshot `hookletPool.keySet()` before iteration so shutdown hooks can mutate the pool without concurrent-modification risk
* `InternalPomProperties`

  * replace `printStackTrace()` with Chronicle debug logging when pom properties cannot be loaded
* `OS`

  * collapse the touched `java.net` imports into a wildcard to match the newly referenced networking exception types
* `ClassUtil`

  * remove a redundant inner `try/catch` around recursive superclass method lookup
* `StaticEnumClass` / `DynamicEnumClass`

  * remove duplicated Javadoc blocks

## Behaviour/impact notes

The main intent of this PR is to make exception boundaries more precise, not to change steady-state behaviour.

The places with real behavioural or diagnostic impact are:

* ignored mock proxies now return primitive defaults instead of `null`
* shutdown hook execution is resilient to pool mutation during shutdown
* background resource release now uses `Jvm.pause(...)` and exits its wait loop on interrupt
* pom-property load failures now go to Chronicle debug logging instead of stderr stack traces
* the build `tmp` directory creation path now uses `mkdirs()`
* the two explicit cleanup boundaries now catch `Throwable`, not just `Exception`, to prioritise cleanup containment

## Notes for reviewers

Please focus the review on the catch-boundary choices rather than the small housekeeping changes.

Areas I’d especially appreciate eyes on:

1. `Jvm` optional platform/reflection paths, especially:

   * direct-memory / reserved-memory fallback handling
   * the do-not-close-on-interrupt reflection paths
   * classpath URL handling
2. The two deliberate `catch (Throwable)` cleanup boundaries in:

   * `AbstractReferenceCounted`
   * `CleaningThreadLocal`
3. The bundled behavioural fixes in:

   * `Mocker`
   * `PriorityHook`
   * `BackgroundResourceReleaser`
   * `InternalPomProperties`
   * `OS`
4. `ObjectUtils.newInstanceOrNull(...)`, which intentionally keeps a broad catch because its contract is still “return null on failure”

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/OS.java
+++ b/src/main/java/net/openhft/chronicle/core/OS.java
@@ -122,7 +122,7 @@ public final class OS {
                 && new File(tmp).isDirectory()
                 && new File(tmp).canWrite())
             return tmp;
-        new File("tmp").mkdirs();
+        new File("tmp").mkdir();
         return "tmp";
     }
 
--- a/src/main/java/net/openhft/chronicle/core/OS.java
+++ b/src/main/java/net/openhft/chronicle/core/OS.java
@@ -18,10 +18,7 @@ import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.net.DatagramSocket;
-import java.net.InetAddress;
-import java.net.InetSocketAddress;
-import java.net.Socket;
+import java.net.*;
 import java.nio.channels.FileChannel;
 import java.security.SecureRandom;
 import java.util.Scanner;
@@ -117,7 +114,7 @@ public final class OS {
         String target = Jvm.getProperty("project.build.directory");
         if (target != null) {
             final File tmp = new File(target, "tmp");
-            tmp.mkdir();
+            tmp.mkdirs();
             return tmp.getPath();
         }
         final String tmp = Jvm.getProperty("java.io.tmpdir");
@@ -702,7 +699,7 @@ public final class OS {
         static String getIpAddressByLocalHost() {
             try {
                 return InetAddress.getLocalHost().getHostAddress();
-            } catch (Throwable e) {
+            } catch (UnknownHostException e) {
                 return "";
             }
         }
@@ -714,7 +711,7 @@ public final class OS {
                     socket.connect(new InetSocketAddress("google.com", 80));
                     return socket.getLocalAddress().getHostAddress();
                 }
-            } catch (Throwable e) {
+            } catch (IOException e) {
                 return "";
             }
         }
@@ -726,7 +723,7 @@ public final class OS {
                     socket.connect(InetAddress.getByName(GOOGLE_DNS), 10002);
                     return socket.getLocalAddress().getHostAddress();
                 }
-            } catch (Throwable e) {
+            } catch (IOException e) {
                 return null;
             }
         }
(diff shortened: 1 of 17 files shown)


src/main/java/net/openhft/chronicle/core/OS.java
```diff
@@ -117,15 +114,15 @@ private static String findTmp0() {
...
+            tmp.mkdirs();
             return tmp.getPath();
         }
         final String tmp = Jvm.getProperty("java.io.tmpdir");
         if (tmp != null
                 && new File(tmp).isDirectory()
                 && new File(tmp).canWrite())
             return tmp;
-        new File("tmp").mkdirs();
+        new File("tmp").mkdir();
```
tgd, line 130: Is this an intentional change? The code won't create intermediate directories any more

tgd (changes requested):

src/main/java/net/openhft/chronicle/core/OS.java
peter-lawrey: The creation of "tmp" shouldn't result in intermediate directories, the use of mkdirs is confusing.

tgd (approved):

PR #855, opened 2026-06-03
Revert "CORE-62: Ensure StringBuilder handles Latin1 encoding in setL…

benbonavia:
Reverts changes of CORE-62 to unblock build all


benbonavia:
fixed forward in Bytes

PR #856, opened 2026-06-03
CORE-67: warn when selected ByteBuffer cleaner cannot free direct memory

james-mcsherry:
### Summary
  `CleanerServiceLocator` selected a cleaner with no indication of which one or whether it worked. A cleaner that cannot free direct memory was selected silently, leaking buffers until GC.

  ### Changes
  - After selection, `CleanerServiceLocator` probes the cleaner once: allocate a direct buffer, `clean()` it, and compare `Jvm.usedDirectMemory()`.
    - frees memory → `debug` naming the cleaner
    - frees nothing, or `Impact.UNAVAILABLE` → `Jvm.warn()` that buffers will leak, naming the cleaner
    - accounting unavailable → silent
  - Self-reported `Impact` is not trusted (a no-op cleaner can report `NO_IMPACT` yet free nothing).
  - Probe is diagnostic only: wrapped so it cannot throw or change the returned cleaner; runs once behind memoisation.
  - `IOToolsTest` ignores the expected warning (the only `CoreTestCommon` test that hits the cleaner path; test classpath registers the no-op `AllowedCleaner`).

  ### Tests
  - New `CleanerServiceLocatorTest` cases: no-op `NO_IMPACT` (incl. dedicated `AllowedCleaner`), no-op `SOME_IMPACT`, `UNAVAILABLE`, genuinely-freeing, real JDK9

### DIFF
--- a/src/main/java/net/openhft/chronicle/core/cleaner/CleanerServiceLocator.java
+++ b/src/main/java/net/openhft/chronicle/core/cleaner/CleanerServiceLocator.java
@@ -136,6 +136,6 @@ public final class CleanerServiceLocator {
 
     private static void warnLeakingCleaner(final String name) {
         Jvm.warn().on(CleanerServiceLocator.class, "Selected ByteBuffer cleaner " + name +
-                " does not free direct memory; direct ByteBuffers will leak until GC.");
+                " does not free direct memory.");
     }
 }
--- a/src/main/java/net/openhft/chronicle/core/cleaner/CleanerServiceLocator.java
+++ b/src/main/java/net/openhft/chronicle/core/cleaner/CleanerServiceLocator.java
@@ -8,8 +8,10 @@ import net.openhft.chronicle.core.annotation.TargetMajorVersion;
 import net.openhft.chronicle.core.internal.cleaner.ReflectionBasedByteBufferCleanerService;
 import net.openhft.chronicle.core.cleaner.spi.ByteBufferCleanerService;
 
+import java.nio.ByteBuffer;
 import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
+import java.util.function.LongSupplier;
 
 /**
  * A utility class to locate the appropriate {@link ByteBufferCleanerService} implementation.
@@ -65,6 +67,8 @@ public final class CleanerServiceLocator {
                 Jvm.warn().on(CleanerServiceLocator.class, "Unable to find suitable cleaner service, falling back to using reflection");
             }
 
+            verifySelectedCleaner(cleanerService);
+
             instance = cleanerService;
             initialised = true;
         }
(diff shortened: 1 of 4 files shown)


src/main/java/net/openhft/chronicle/core/cleaner/CleanerServiceLocator.java
```diff
@@ -87,4 +91,52 @@ private static boolean isAllowedInThisMajorVersion(final ByteBufferCleanerServic
...
+
+            if (afterClean < afterAllocation) {
+                Jvm.debug().on(CleanerServiceLocator.class, "Selected ByteBuffer cleaner: " + name +
+                        " (impact=" + svc.impact() + ", verified to free direct memory)");
+            } else {
+                warnLeakingCleaner(name);
+            }
+        } catch (Throwable t) {
+            // diagnostic only — must not affect selection
+            Jvm.debug().on(CleanerServiceLocator.class, "Could not verify ByteBuffer cleaner " + name, t);
```
peter-lawrey: I would make this an error as something pretty bad must have happened for this to not work

peter-lawrey (changes requested):

src/main/java/net/openhft/chronicle/core/cleaner/CleanerServiceLocator.java
james-mcsherry: Thanks @peter-lawrey - I have made the change

tgd:
Hi @james-mcsherry I get this failure when running `mvn clean install` locally:

```
[ERROR] Failed to execute goal com.mycila:license-maven-plugin:4.6:check (license-check) on project chronicle-core: Some files do not have the expected license header. Run license:format to update them. -> [Help 1]
```

james-mcsherry:
Hi @tgd - weirdly I don't get that error on mine? I ran mvn license:format before committing

src/main/java/net/openhft/chronicle/core/cleaner/CleanerServiceLocator.java
```diff
@@ -87,4 +91,51 @@ private static boolean isAllowedInThisMajorVersion(final ByteBufferCleanerServic
...
+                warnLeakingCleaner(name);
+            }
+        } catch (Throwable t) { // NOSONAR — diagnostic probe must never destabilise selection, incl. Errors
+            Jvm.error().on(CleanerServiceLocator.class, "Could not verify ByteBuffer cleaner " + name, t);
+        }
+    }
+
+    private static void warnLeakingCleaner(final String name) {
+        Jvm.warn().on(CleanerServiceLocator.class, "Selected ByteBuffer cleaner " + name +
+                " does not free direct memory; direct ByteBuffers will leak until GC.");
```
benbonavia: This last line in the warn is slightly misleading in my opinion. 

To me, this implies that my application will be ok and all will be taken care of by the GC when the memory gets to the threshold. However, this is not the case, the GC only monitors on-heap memory and the ByteBuffer heap objects are so small in comparison to the size of the direct memory they create, so it is very possible that an out-of-memory exception can be thrown way before the GC will be invoked. 

So perhaps the last sentence could be removed?

james-mcsherry: Good point - fixed!

peter-lawrey (approved):

